PR #19, opened 2018-08-16
Tweak style to align with kubernetes.io

marcoceppi:
I know style is subjective and this is merely an opinion. I'm open to feedback and tweaks, but figured I'd give it a stab.

# Changes

* Uses dark grey as secondary color
* Update blue to be more Kubernetes blue
* Remove superfluous iconography
* Remove drop shadows
* Remove animations


static/css/overrides.css
```diff
@@ -0,0 +1,75 @@
...
+
+/* Arrow if nav has children */
+#sidebar ul li .category-icon {
+  color: #fff;
+}
+
+/* Sidebar */
+#sidebar {
+  background-color: #333;
+  background-image: url('https://kubernetes.io/images/texture.png');
```
mrbobbytables: My only suggestion would be to pull down the image and put it in the local `static/images` directory. That way if the other site is down or the file location changes the contributor site itself would not be impacted.

marcoceppi: True, I meant to do so but missed that in my final pass. I'll put it in `static/`

castrojo:
/approve

Thanks, and welcome @marcoceppi !

PR #33, opened 2018-10-30
Add Calendar to the sidebar

castrojo:
The calendar is important enough to top-level.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -20,5 +20,7 @@ outputs:
 
 menu:
   shortcuts:
+  - name: Calendar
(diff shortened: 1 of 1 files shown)


config.yaml
```diff
@@ -20,5 +20,7 @@ outputs:
 
 menu:
```
mrbobbytables: It should look like this to ensure it does not open in a new page.

```yaml
menu:
  main:
  - name: Community Calendar
    url: /calendar/
  shortcuts:
  - name: Glossary
    url: https://kubernetes.io/docs/reference/glossary/?community=true
```

PR #34, opened 2018-10-30
Bob is doing a bunch of work here, he should be an OWNER

castrojo:
/cc @parispittman @idvortskyi @cblecker


parispittman:
/lgtm

I'd approve but just want to know from @cblecker if there is anything else we need to update before I do. [I don't think so but never have done an OWNERs change in k-sigs before]

PR #76, opened 2019-10-04
Update OWNERS.

mrbobbytables:
Updates OWNERS with other owners within the contributor documentation subproject.

/cc @guineveresaenger @tpepper @justaugustus @kbarnard10


guineveresaenger:
+1 for my part in this. ❤️🙇‍♀️

However, that’s a lot of subtractions - feeling like we need to add more folks.

mrbobbytables:
> However, that’s a lot of subtractions - feeling like we need to add more folks.

It was a good chunk of folk, but those were also carryovers from when the repo was first created. Pretty much all the PRs/Reviews from either @castrojo or I.

PR #82, opened 2019-11-07
Remove extra isCurrentPage

whs-dot-hk:
Looks like this line is not use anymore

what do you think?

Closes https://github.com/kubernetes-sigs/contributor-site/issues/80


mrbobbytables:
Thanks @whs-dot-hk :)
/lgtm
/approve

PR #91, opened 2020-01-13
fix: External icon was not setted with rel path

vishakhanihore:
- External icon was loading at other pages


mrbobbytables:
verified it works 👍 
/lgtm
/approve

Issue #71, opened 2019-08-02
coc: display coc


cblecker:
The Code of Conduct link points directly to the code-of-conduct.md in the community repo. Is it possible to render that file in Hugo?

PR #85, opened 2019-12-29
Improvising ReadME and Code of Conduct section

vishakhanihore:
Fixes #71 
- This PR also revamps the READme Section for setting up the project locally without a docker container.

PR #92, opened 2020-01-19
Support for non-kubernetes org external sources

mrbobbytables:
This adds support for non-kubernetes org sources.
external-sources now follows the pattern of:
`external-sources/<org|person>/<repo>`

Essentially it should mimic the github url structure. The only additional item that is needed when adding a new org/person is the [regex](https://github.com/mrbobbytables/contributor-site/blob/support-more-external-sources/hack/gen-content.sh#L183) that detects if the links are linking to an internal source or an external source will need to be updated. 
```
grep -q -i -E "https?:\/\/((sigs|git)\.k8s\.io|(www\.)?github\.com\/(kubernetes(-(client|csi|incubator|sigs))?|cncf))"`
```
For most it should be just appending unless there is some shortener similar to git.k8s.io used by the other org.

---

The CoC page is now generated as the first external non-kubernetes source to be brought in.

---

As a last note -- the gen-content.sh script has grown significantly than it's inception and its likely time to be replaced by something more sustainable such as go or python. I'll open an issue for further discussion

---

fixes: #71 
supersedes: #85

/cc @vishakhanihore @nimbinatus
/assign @castrojo


content/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Files synced to the root of the content directroy should be added to the
```
vishakhanihore: *directory

hack/gen-content.sh
```diff
@@ -176,14 +178,16 @@ gen_link() {
 
   # If it was previously an "external" link but now local to the contributor site
   # update the link by trimming the url portion.
-  if echo "$generated_link" | grep -q -i -E "[http|https]://(git.k8s.io|(www\.)?github.com/kubernetes)"; then
+  # Detection is a  problem for non kubernetes repos.
+  # TODO: Support other external repos
+  if echo "$generated_link" | grep -q -i -E "https?:\/\/((sigs|git)\.k8s\.io|(www\.)?github\.com\/(kubernetes(-(client|csi|incubator|sigs))?|cncf))"; then
```
vishakhanihore: @mrbobbytables Can we add a comment here regarding the regex as you mentioned in the issue ?

hack/gen-content.sh
```diff
@@ -235,29 +261,39 @@ main() {
   local dsts=() # array of destinations for the content to be synced to
 
   # Files within the EXTERNAL_SOURCES directory should be csv formatted with the
-  # name of the file being the kubernetes repo name (e.g. community), and the
-  # content being the path to the content to be synced within the repo to the
-  # to the destination within the HUGO contenet directory. 
+  # directory being the GitHub org and name of the file being the repo name
+  # (e.g. kubernetes/community), and the  content being the path to the content
+  # to be synced within the repo to the to the destination within the HUGO
+  # contenet directory.
```
vishakhanihore: Content

mrbobbytables:
@vishakhanihore comments have been addressed 👍 please give it another look when you have a chance.

PR #95, opened 2020-01-20
feat: Upgrading to goldmark markdown renderer

### LINKED ISSUE #94: Update site to use Hugo v0.60.0+
Hugo [v0.60.0](https://github.com/gohugoio/hugo/releases/tag/v0.60.0) introduced a new markdown renderer called [goldmark](https://github.com/yuin/goldmark/). This new renderer is much closer to native GitHub flavored markdown which should eliminate some of the idiosyncrasies that have been seen with the current [BlackFriday](https://github.com/russross/blackfriday) based rendering engine. 

Goldmark and several other changes in v0.60.0+ may potentially introduce some breaking changes with the current site and should be evaluated with care.

/kind feature

vishakhanihore:
Fixes #94

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@ publish = "public"
 command = "make production-build"
 
 [build.environment]
-HUGO_VERSION = "0.55.0"
+HUGO_VERSION = "0.62.0"
 
 [context.deploy-preview]
 command = "make preview-build"
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 
 DOCKER			?= docker
 DOCKER_RUN		:= $(DOCKER) run --rm -it -v $(CURDIR):/src
-HUGO_VERSION		:= 0.55.6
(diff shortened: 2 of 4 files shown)


vishakhanihore:
@mrbobbytables I didn't see any breaking changes in 0.61, Also if there is something more in upgrading except changing the toml file. Please let me know.

netlify.toml
```diff
@@ -3,7 +3,7 @@ publish = "public"
 command = "make production-build"
 
 [build.environment]
-HUGO_VERSION = "0.55.0"
+HUGO_VERSION = "0.62.0"
```
mrbobbytables: It also needs to be updated within the Makefile. It might be worth doing something similar to the [kubernetes/website Makefile](https://github.com/kubernetes/website/blob/master/Makefile#L2) and have `netlify.toml` be the source of truth. That way we don't need to update it in two places.

vishakhanihore:
> This is a little thing, but would you mind breaking out the changes to the moc docs from the commit adding goldmark? It'd just ensure a clean history :x

I will take a closer look at changelogs and see if there is a chance of potential breakdown for app.

mrbobbytables:
Thanks @vishakhanihore :) 

/approve
/assign @castrojo

PR #97, opened 2020-01-21
fix: Merging the .gitignore with one present in root

vishakhanihore:
- There are two .gitignore in our repository. Merging them into one

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,9 @@ Session.vim
 public/
 resources/
 
+# Files synced to the root of the content directory should be added to the ignore list.
+content/code-of-conduct.md
+
 # Temp directory for processing content
 _tmp/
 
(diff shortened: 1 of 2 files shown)


mrbobbytables:
My thought behind putting it in the content directory was to act as a signal that there is some things within there that should be ignored. I'm fine with bumping it back up to the root `.gitignore` though.

vishakhanihore:
> My thought behind putting it in the content directory was to act as a signal that there is some things within there that should be ignored. I'm fine with bumping it back up to the root `.gitignore` though.

I think a single `.gitignore` with an explanation in comment should do the work. But if having two would ease the flow, I have no issues in closing.
What do you suggest ?

mrbobbytables:
¯\\_(ツ)_/¯
/approve

PR #106, opened 2020-01-30
Import release timeline doc

### LINKED ISSUE #2: Need a release dashboard
We should help the release team by making contributor.k8s.io/release a useful place. 

Currently the release team uses a bit.ly link to point people to the right place. http://bit.ly/k8s112-release-info

Ideas for things to include on this page (please include more!):

- A banner up top "We are currently in: $STATE". Freeze, slush, etc. with a link to what each of those terms mean.
- A clever way to integrate the important release dates and make them obvious
- Links that show the latest development release, stable release, and older supported release version numbers, clickable to allow for driving down into release info. 
- List and contact information for each release team member for that cycle.
- Integrate the [k8s-release](https://discuss.kubernetes.io/tags/k8s-release) tag so users can see the latest announcements from the release team. 
- Shortcut links to appropriate devstats charts the team deems important. 

For now, we can just make the URL autoforward to the right place on github so we can socialize it.

mrbobbytables:
This both imports the release docs and fixes a bug when importing and renaming a README document when the destination is intended to be renamed.

"Previously the script made a false assumption that source and destination filenames would be the same UNLESS the file was a readme. This caused an issue when importing a singular source readme with a different name once imported."

fixes #2 
/cc @vishakhanihore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ resources/
 
 # Files synced to the root of the content directory should be added to the ignore list.
 content/code-of-conduct.md
+content/release.md
 
 # Temp directory for processing content
 _tmp/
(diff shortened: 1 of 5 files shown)


vishakhanihore:
I can't see the release subtab in netlify deploy ?

mrbobbytables:
/hold
Looks like theres an issue. Locally it seems to render fine, but not on netlify. Need to solve that before this can be merged

mrbobbytables:
Okay, should be fixed. Please go take a look 👍

vishakhanihore:
/cancel hold 
/lgtm
Perfect 🎉

Issue #105, opened 2020-01-30
bug: Some URLs are not being generated correctly


mrbobbytables:
Relative source URLs referencing external content are not being generated correctly.

Example:
CLA link from here:
https://kubernetes-contributor.netlify.com/guide/contributor-cheatsheet/#signing-the-cla

Original:
```
[cla]: /CLA.md#how-do-i-sign
```

After being imported:
```
[cla]: /CLA.md#how-do-i-sign
```

It should be expanded to reference the full github url below:
```
[cla]: https://github.com/kubernetes/community/blob/master/CLA.md#how-do-i-sign
```

/kind bug
/priority important-soon

/assign

vishakhanihore:
> Original:
> 
> ```
> [cla]: /CLA.md#how-do-i-sign
> ```
> 
> After being imported:
> 
> ```
> [cla]: /CLA.md#how-do-i-sign
> ```

@mrbobbytables What is the bug here? Do we need to include the reference to the GitHub URL in the URL of the content page?

mrbobbytables:
That 404's when viewing it on the contributor site. The link generator should update the link so that it works - It'd link to the GithHub page by updating it to be this:
https://github.com/kubernetes/community/blob/master/CLA.md

It does that for MANY URLs for example, in contributing.md the e2e test url is this:
```
[document detailing e2e testing](/contributors/devel/sig-testing/e2e-tests.md)
```
That link would break on the contributor site because it is a file reference. The process_content function will expand it to this:
```
[document detailing e2e testing](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md)
```
For reference:
https://github.com/kubernetes-sigs/contributor-site/blob/fe36c430d92ed0a81cb85e154fa092f55912a3be/hack/gen-content.sh#L68-L89

I do need to update it to say it does not use the shortner anymore, but will update it with the github.com address

PR #110, opened 2020-02-06
Fix reference link expansion

mrbobbytables:
fixes: #105 

TL;DR had an extra `|` in the regex when there shouldn't have been 🤦‍♂ 

/cc @vishakhanihore


vishakhanihore:
/lgtm 
Looks Great !!

PR #122, opened 2020-04-05
 Docsy conversion

mrbobbytables:
preview of docsy build

/hold


kushthedude:
@mrbobbytables I noticed some issues while setting up branch :
- I think there is a need to commit `yarn.lock` too as there is only package.json.
- Also while setting up the branch I faced the following Hugo Issue which I think should be addressed in PR comments or Project Documentation https://github.com/gohugoio/hugo/issues/6274 
```sh
Building sites … ERROR 2020/04/16 19:47:40 Transformation failed: SCSS processing failed: file "stdin", line 6, col 1: File to import not found or unreadable: ../vendor/bootstrap/scss/bootstrap.
```
- Also since we are now using package.json , will it be better to use bootstrap as a node_module ?

mrbobbytables:
> I think there is a need to commit yarn.lock too as there is only package.json.

I was avoiding yarn to mostly stay inline with docsy and the main kubernetes website that stick to npm. I should probably go ahead and commit the package-lock.json though.

> We need to use following workaround to fix this(I am not sure if this is correct way to approach it)
> Building sites … ERROR 2020/04/16 19:47:40 Transformation failed: SCSS processing failed: file "stdin", line 6, col 1: File to import not found or unreadable: ../vendor/bootstrap/scss/bootstrap.

🤔 Are you sure? It's being built by netlify without the module change.

kushthedude:
> I was avoiding yarn to mostly stay inline with docsy and the main kubernetes website that stick to npm. I should probably go ahead and commit the package-lock.json though.

Yeah, any node engine would work. I installed using yarn.

> > We need to use following workaround to fix this(I am not sure if this is correct way to approach it)
> > Building sites … ERROR 2020/04/16 19:47:40 Transformation failed: SCSS processing failed: file "stdin", line 6, col 1: File to import not found or unreadable: ../vendor/bootstrap/scss/bootstrap.
> 
> 🤔 Are you sure? It's being built by netlify without the module change.

Yes, The following is the error log from my machine & the issue is still open on their official repository. Maybe there could be some issue of OS. I am on macOS Catalina

mrbobbytables:
> Yes, The following is the error log from my machine & the issue is still open on their official repository. Maybe there could be some issue of OS. I am on macOS Catalina

I'm on Mojave with Hugo `v0.68.3/extended` installed.
I also have the postCSS modules installed for docsy:
ref: https://www.docsy.dev/docs/getting-started/#install-postcss

kushthedude:
> > Yes, The following is the error log from my machine & the issue is still open on their official repository. Maybe there could be some issue of OS. I am on macOS Catalina
> 
> I'm on Mojave with Hugo `v0.68.3/extended` installed.
> I also have the postCSS modules installed for docsy:
> ref: https://www.docsy.dev/docs/getting-started/#install-postcss

Same issue. BTW you have already defined both `autoprefixer & postcss` so npm i should have installed them already.

mrbobbytables:
Do you have the extended version of hugo installed?

kushthedude:
> Do you have the extended version of hugo installed?

Yes, `Hugo Static Site Generator v0.68.3/extended darwin/amd64`

mrbobbytables:
Added the package-lock -- I did have another thought -- did you do this to grab the docsy theme and it's dependencies?
```
git submodule update --init --recursive
```

the recursive command should grab the 2 git repos that it itself needs
ref: https://github.com/google/docsy/tree/master/assets/vendor

kushthedude:
Oops, I just did it once ! I didn't use arg twice while submodule init.

kushthedude:
> git submodule update --init --recursive

It worked! 
BTW I had another thing in mind since docsy theme is not ready why not we move the development of the new site to a new branch named `docsyVersion` or anything else? We can merge the branch into master when the site is ready to be live.

mrbobbytables:
/hold cancel

There are still some issues, but this is in a good enough state to merge 👍

PR #131, opened 2020-05-20
Fix svg render bug

mrbobbytables:
There is a bug in hugo and the upstream minify library that causes svgs
to not be rendered correctly. This has been fixed in the upstream
library (https://github.com/tdewolff/minify) as of minify 2.7.3. This
can be reverted or undone once Hugo upgrades to using that version or
better.

/kind bug
ref: https://github.com/kubernetes-sigs/contributor-site/issues/129


jeefy:
Weird.

/lgtm

PR #164, opened 2020-09-11
feat: add netlify rebuild periodic job

kushthedude:
Signed-off-by: Kush Trivedi <kushthedude@gmail.com>

### DIFF
--- a/.github/workflows/netlify-periodic-build.yml
+++ b/.github/workflows/netlify-periodic-build.yml
@@ -1,0 +1,12 @@
+name: Trigger netlify build
+on:
+  schedule:
+  - cron: '30 3 * * *'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Trigger webhook on netlify
+      run: curl -s -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
+      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/netlify-periodic-build.yml
```diff
@@ -0,0 +1,12 @@
+name: Trigger netlify build
+on:
+  schedule:
+  - cron: '30 3 * * *'
```
mrbobbytables, line 4:
```suggestion
  - cron: '0 */8 * * *'
```
I was thinking about this - I think we can get away with it firing to update every 8 hours. That should be 3 times a day and in general more than enough (at least for now XD). What do you think?

kushthedude: I agree but it depends upon the netlify plan as netlify started plan allows 300 build minutes/month and if I consider a build to last to take 3.5-4 minutes, that will almost exhaust 300 build minutes with just periodic deploys in a month.

mrbobbytables: Good point, we're on enterprise but I completely forgot that netlify build minutes cost $$$$.

We can stick with daily at 3:30am :)

PR #171, opened 2020-10-27
Add Google Custom Search

adamkane13:
This uses Google custom search for the Docsy search engine. Code was generated with contributors@kubernetes.io account for centralized maintenance.


castrojo:
/lgtm

We followed the directions from google to do this, note that they say that it can take up to a few days for the results to start to show up.

mrbobbytables:
We will need to push it live to actually test. Will approve and if there is a problem, revert.
/approve

PR #181, opened 2020-12-01
Restore website minification via hugo minify

fntlnz:
Hi 🤗 !!

It was disabled in #131 because hugo is not using 1,
however, the hugo project has been trying to update this but it always
introduces regressions and they are having regressions with the JS
minifier.

- https://github.com/gohugoio/hugo/pull/7701
- https://github.com/gohugoio/hugo/issues/7792

I felt like it was a good idea to restore the wide minification and only disable the svg one instead of disabling it completely.

I think this could also be contributed to docsy directly but I thought that doing it here would've been equally good since it's a temporary workaround. 

Once hugo has `tdewolff/minify >= 2.7.3` we can just remove the `navbar.html` partial, I wrote a notice in the file about that too.

Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,8 +106,7 @@ production-build: ## Builds the production site (this command used only by Netli
 		--verbose \
 		--buildFuture \
 		--ignoreCache \
-
-		# --minify -Add back in when hugo upgrades to minify 2.7.3 or greater
+		--minify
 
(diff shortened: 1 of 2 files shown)


mrbobbytables:
Thanks for fixing that :D It's been a long standing issue.
/lgtm
/approve
/pony yay

PR #185, opened 2020-12-07
📝 Add a guide on internet board games for the CC docs.

coderanger:
First pass from the google doc.

### DIFF
--- a/content/en/events/kcc2020/activities.md
+++ b/content/en/events/kcc2020/activities.md
@@ -12,12 +12,27 @@ type: docs
   - [Kubernetes Contributor Awards](#kubernetes-contributor-awards)
   - [Most Extreme Kubernetes Challenge - DevOps Party Game](#most-extreme-kubernetes-challenge---devops-party-game)
   - [The Great Cloud Native Bake Off](#the-great-cloud-native-bake-off)
-- [Games](#games)
+- [Party Games](#party-games)
   - [Among Us](#among-us)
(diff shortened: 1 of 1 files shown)


content/en/events/kcc2020/board-games.md
```diff
@@ -0,0 +1,140 @@
...
+
+Just because we are internet-partying doesn’t mean we need to skimp on the board games! This is not an exhaustive list and if you have any questions please join us in the board-games channel on Discord!
+
+## Web
+
+All of these only require a web browser to get started.
+
+### Board Game Arena
+
+URL: https://boardgamearena.com/
```
castrojo, line 16: These aren't showing up as bullets/linefeeds in the markdown preview so they're all on the same line. Same for all the other games.

mrbobbytables:
This is going back a bit on what I said in slack XD but after thinking about it more - I think it would probably be better to fold under our events and activities with board games as an H2 and then the rest under it as H3. 

What do you think?

coderanger:
> This is going back a bit on what I said in slack XD but after thinking about it more - I think it would probably be better to fold under our events and activities with board games as an H2 and then the rest under it as H3.
> 
> What do you think?

Defer to you on IA, just worried that's getting to be a very long doc :)

coderanger:
Reworked to be part of the activities page, along with reformatting to better support the new content.

jeefy:
/lgtm
:taco:

PR #192, opened 2020-12-17
fix few minor typos

rajatjindal:
while deprecating docker-server with container-serve (or container-server), there was some inconsistency.  

e.g. in container-targets, it refers to container-server, while other places including the target itself, its used as container-serve.

```Makefile
container-targets: container-image container-gen-content container-render container-server
```

```Makefile
docker-server:
	@echo -e "**** The use of docker-render is deprecated. Use container-serve instead. ****" 1>&2
	$(MAKE) container-serve

container-serve: ## Run Hugo locally within a container, available at http://localhost:1313/
```

Please let me know if we want to keep it container-serve (if the change was intentional) or container-server (consistent with outgoing docker-server).

thanks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,10 +71,10 @@ container-render: ## Build the site using Hugo within a container (equiv to rend
 	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hugo --verbose --ignoreCache --minify
 
 docker-server:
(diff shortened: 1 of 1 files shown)


rajatjindal:
requesting review from @castrojo @mrbobbytables

mrbobbytables:
Thanks!
/lgtm
/approve

PR #206, opened 2021-02-20
added path for mentorship.md

dineshkumar181094:


dineshkumar181094:
https://www.kubernetes.dev/mentoring/README.md
error 404

idvoretskyi (approved):
/lgtm

mrbobbytables:
/lgtm cancel

Adding pages is not that easy. =\ It requires the original page to be prepped with hugo metadata and the site on this side configured with a section, added to git ignore etc
Right now all we want is the MoC page because it's more of an event. We may add the mentoring category in the future, but it'd likely fall under a new header.

/close

dineshkumar181094:
@mrbobbytables  thanks for the quick response just attending a meet and found the speaker was unable to get this page. just thought of fixing this.

PR #216, opened 2021-04-06
Add dotglob for globbing in hack/gen-content.sh

dholbach:
We are using `hack/gen-content.sh` for <https://fluxcd.io> as well and recently added `dotglob` for globbing to allow us pulling in <https://github.com/fluxcd/.github/> as you can see in <https://github.com/fluxcd/website/blob/main/external-sources/fluxcd/.github>.


idvoretskyi (approved):
/lgtm

thanks @dholbach!

mrbobbytables:
/approve
Thanks!

PR #219, opened 2021-05-10
Bump release info page

mrbobbytables:
/assign @savitharaghunathan @annajung @erismaster @divya-mohan0209 @kikisdeliveryservice

### DIFF
--- a/external-sources/kubernetes/sig-release
+++ b/external-sources/kubernetes/sig-release
@@ -1,1 +1,1 @@
-"/releases/release-1.21/README.md","/resources/release/index.md"
+"/releases/release-1.22/README.md","/resources/release/index.md"


mrbobbytables:
/hold cancel
I think this is fine, we've done it for the past several releases, but missed updating this page earlier in the release cycle =/
Probably a note for the retro to make this a task once the readme is created.

divya-mohan0209:
> /hold cancel
> I think this is fine, we've done it for the past several releases, but missed updating this page earlier in the release cycle =/
> Probably a note for the retro to make this a task once the readme is created.

@mrbobbytables : Would the RT Lead/Lead Shadows be the ones with the AI of doing this typically?

mrbobbytables:
I think that'd be best. They'll know as soon as that readme is in a good enough state to publish

PR #224, opened 2021-06-28
migrate Release Managers page

neha-viswanathan:
Related to https://github.com/kubernetes/website/issues/28055

### DIFF
--- a/content/en/releases/release-managers.md
+++ b/content/en/releases/release-managers.md
@@ -1,0 +1,214 @@
+---
+title: Release Managers
+---
+
+"Release Managers" is an umbrella term that encompasses the set of Kubernetes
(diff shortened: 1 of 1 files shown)


mrbobbytables (commented):
/hold

This looks like it's adding a page and then removing it?

If more release related info is going to be added we can create a section under docs for separately

mrbobbytables:
@neha-viswanathan is there any plan to move forward with this? If so it needs to be in a separate section, it can't just go at the root the docs repo and it needs the appropriate metadata added to the top of the doc.

/lifecycle rotten

mrbobbytables:
There has been no movement on this. Going to go ahead and close.
/close

PR #230, opened 2021-07-22
Update links to use git.k8s.io where possible

spiffxp:
Fewer URI's to update when repos decide to rename their default branches.

Generated via:

    for f in $(grep -l -r 'github.com/kubernetes'); do \
      sed -i.bak -e 's!https://github.com/kubernetes/([^/]*)/(tree|blob)/master/!https://git.k8s.io/\1/!g'; \
    done
    find . -name '*.bak' -exec rm {} \;
    git add .

Also snuck in a dead link fix to follow https://github.com/kubernetes/k8s.io/pull/2383 (which honestly is what originally motivated me to look at this repo)

If retroactive blog post updates for link fixes aren't allowed, my apologies


hack/gen-content.sh
```diff
@@ -82,8 +82,8 @@ find_md_files() {
 #   Links:
 #   ./bug-bounty.md -> /guide/bug-bounty
 #   contributor-cheatsheet/README.md -> /guide/contributor-cheatsheet
-#   ../../sig-list.md -> https://github.com/kubernetes/community/blob/master/sig-list.md
-#   /contributors/devel/README.md -> https://github.com/kubernetes/community/blob/master/contributors/devel/README.md
+#   ../../sig-list.md -> https://git.k8s.io/community/sig-list.md
+#   /contributors/devel/README.md -> https://git.k8s.io/community/contributors/devel/README.md
```
mrbobbytables, line 86: One small thing, this is an example the link expander. This should probably go back or what it was previously as it expands to the GitHub link independent of the org and git.k8s.io afaik is only for the kubernetes org

Otherwise lgtm 👍 


(could also be done in a follow up PR)

spiffxp: OK I reverted this change in a separate commit explaining why

spiffxp:
Updated, PTAL

PR #270, opened 2021-12-14
Remove BakeOff from the description, as it's cancelled.

jberkus:
Unfortunately we had insufficient participation in the BakeOff and it's now off the schedule.

It will be replaced by a culinary event on Friday to be added by the III team.


mrbobbytables (commented):
Can you remove the entry from the table of contents too?^^;;;;

jberkus:
ah, point.  Done.

PR #222, opened 2021-06-08
All Enhancement proposals in single webpage | KEPs

shekhar-rajak:
This PR is for https://github.com/kubernetes/enhancements/issues/2095 and we want a KEP webpage to find out about specific kep and its metadata easily.

1. keps.json is generated from enhancements repo using gen-content.sh
2. sigs.yaml is fetched from community/sigs.yaml file

Refer: https://github.com/kubernetes/website/pull/28295

PR #324, opened 2022-07-27
Add page to list all KEPs

palnabarun:
Part of https://github.com/kubernetes/enhancements/issues/2095

The proposed change now uses [Datatables](https://datatables.net/) for table search, pagination and sorting. No additional custom JS is required. Also, there is no change done to the content generator hack script.

Here is how the page looks:
image

Supersedes #222 (Thanks, @shekhar-rajak for the initial work 🎉 )

> Note: A lot of future improvements can be done on this page. This is the MVP that we should get to first.

### DIFF
--- a/.github/workflows/netlify-periodic-build.yml
+++ b/.github/workflows/netlify-periodic-build.yml
@@ -1,7 +1,9 @@
 name: Trigger netlify build
 on:
   schedule:
-  - cron: '30 3 * * *'
+  # The build is triggered every 6 hours so that the KEP page remains fresh.
+  # TODO(future): Revisit the cadence.
+  - cron: '0 */6 * * *'
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/content/en/resources/keps/_index.md
+++ b/content/en/resources/keps/_index.md
@@ -1,0 +1,22 @@
+---
+linktitle: Enhancements
+title: Kubernetes Enhancement Proposals (KEPs)
+description: |
+  List of Kubernetes enhancements
+type: docs
+---
+
+These data come from the [kubernetes/enhancements]
+repository on GitHub.
+
+To see the original Enhancement issue, click on the KEP number.  To see a summary of what the KEP does, click on the Title.
+
+{{% alert title="Note" color="info" %}}
+If you notice an empty cell in the table, the KEP may not have updated information.
+Please raise <a href="https://github.com/kubernetes/enhancements/issues">an issue in the kubernetes/enhancemenets</a> repository.
+{{% /alert %}}
+###	KEP List
+
+{{< keps-data >}}
+
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -1,0 +1,25 @@
+{{/*
+    This file is a 1:1 copy of the one in docsy but with adding scripts
+    for enabling Datatables on the KEP page.
+
+    Remove this notice if a better way is found to include the scripts.
+*/}}
+
+{{ with .Site.Params.algolia_docsearch }}
+<!-- scripts for algolia docsearch -->
+{{ end }}
+
+{{ if .HasShortcode "keps-data" }}
+<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.12.1/js/jquery.dataTables.js"></script>
+<script type="text/javascript">
+    $(document).ready(function () {
+        $('#keps-table').DataTable({
+            lengthMenu: [
+                [10, 25, 50, -1],
+                [10, 25, 50, 'All'],
+            ],
+            order: [[2, 'asc'], [0, 'asc']],
+        });
+    });
+</script>
+{{ end }}
--- a/layouts/shortcodes/keps-data.html
+++ b/layouts/shortcodes/keps-data.html
@@ -1,0 +1,86 @@
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.12.1/css/jquery.dataTables.css">
+<style>
+  .kep-milestone {
+    font-weight: bold;
+  }
+</style>
+
+<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.12.1/js/jquery.dataTables.js"></script>
+<script type="text/javascript">
+  $(document).ready( function () {
+    $('#keps-table').DataTable({
+      lengthMenu: [
+        [10, 25, 50, -1],
+        [10, 25, 50, 'All'],
+      ],
+      order: [[2, 'asc'], [0, 'asc']],
+    });
+});
+</script>
+
+<div>
+  <table id="keps-table" class="table table-hover mt-2 col-md-12 table-sm">
+    <thead class="thead-light">
+      <tr>
+        <th scope="col" class="col-md-1">KEP Number</th>
+        <th scope="col" class="col-md-3">Title</th>
+        <th scope="col" class="col-md-2">SIG</th>
+        <th scope="col" class="col-md-2">Author</th>
+        <th scope="col" class="col-md-1">Created</th>
+        <th scope="col" class="col-md-1">Last Updated</th>
+        <th scope="col" class="col-md-1">Milestone</th>
+      </tr>
+    </thead>
+    <tbody>
+      {{range $index, $data := getJSON "https://storage.googleapis.com/k8s-keps/keps.json" }}
+          <tr>
+            <td>
+              <a href="https://features.k8s.io/{{ $data.kepNumber }}">
+                {{ printf "%s" $data.kepNumber }}
+              </a>
+            </td>
+            <td>
+              <a href="https://github.com/kubernetes/enhancements/tree/master/keps/{{$data.owningSig}}/{{$data.name}}#summary">
+                {{ printf "%s" $data.title }}
+              </a>
+            </td>
+            <td>
+              <a href=" https://github.com/kubernetes/community/tree/master/{{ $data.owningSig }}">
+                {{ strings.TrimPrefix "sig-" $data.owningSig | humanize | title }}
+              </a>
+            </td>
+            <td>
+              {{ range $index, $author := $data.authors }}
+              {{- if $index -}}, {{ end -}}
+              <a href="https://github.com/{{ trim $author "@" }}">
+                <span class="github-user">@{{ trim $author "@" }}</span>
+              </a>
+              {{ end }}
+            </td>
+            <td>
+                {{ if $data.creationDate }}
+                  <time datetime="{{ $data.creationDate }}">{{ $data.creationDate }}</time>
+                {{ end }}
+            </td>
+            <td>
+                {{ if $data.lastUpdated }}
+                  <time datetime="{{ $data.lastUpdated }}">{{ $data.lastUpdated }}</time>
+                {{ end }}
+              </span>
+            </td>
+            <td>
+              <ul class="list-unstyled">
+                {{ range $stage, $milestone := $data.milestone }}
+                  {{ if $milestone }}
+                  <li>
+                    {{ $stage }}:<span class="kep-milestone">{{ $milestone }}</span>
+                  </li>
+                  {{ end }}
+                {{ end }}
+              </ul>
+            </td>
+          </tr>
+      {{ end }}
+    </tbody>
+  </table>
+</div>
--- a/layouts/shortcodes/keps-data.html
+++ b/layouts/shortcodes/keps-data.html
@@ -1,22 +1,12 @@
-<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.12.1/css/jquery.dataTables.css">
-<style>
-  .kep-milestone {
-    font-weight: bold;
-  }
-</style>
+{{/*
+  Datatables is used in the page to enrich the list of KEPs.
+  The stylesheets and scripts required are in `partials/hooks/head-end.html`
+  and `partials/hooks/body-end.html` respectively.
 
-<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.12.1/js/jquery.dataTables.js"></script>
-<script type="text/javascript">
-  $(document).ready( function () {
-    $('#keps-table').DataTable({
-      lengthMenu: [
-        [10, 25, 50, -1],
-        [10, 25, 50, 'All'],
-      ],
-      order: [[2, 'asc'], [0, 'asc']],
-    });
-});
-</script>
+  Note: In order to configure Datatable parameters for changing any behavior, edit
+  the `DataTable` initialization in `partials/hooks/body-end.html`
+
+*/}}
 
 <div>
   <table id="keps-table" class="table table-hover mt-2 col-md-12 table-sm">
--- a/layouts/partials/hooks/head-end.html
+++ b/layouts/partials/hooks/head-end.html
@@ -1,0 +1,20 @@
+{{/*
+    This file is a 1:1 copy of the one in docsy but with adding stylesheets
+    for enabling Datatables on the KEP page.
+
+    Remove this notice if a better way is found to include the stylesheets.
+*/}}
+
+{{ with .Site.Params.algolia_docsearch }}
+<!-- stylesheet for algolia docsearch -->
+{{ end }}
+
+{{ if .HasShortcode "keps-data" }}
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.12.1/css/jquery.dataTables.css">
+<style>
+    .kep-milestone {
+        font-weight: bold;
+    }
+
+</style>
+{{ end }}


palnabarun:
/assign @sftim @mrbobbytables 

(since you reviewed #222)

sftim:
Some of the approach in https://github.com/kubernetes/website/pull/35228 might be useful for generating a list of KEPs.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,76 @@
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.12.1/css/jquery.dataTables.css">
+
+<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.12.1/js/jquery.dataTables.js"></script>
+<script type="text/javascript">
+  $(document).ready( function () {
+    $('#keps-table').DataTable();
+});
+</script>
```
sftim, line 8: nit: this could go into a layout, gated behind a `HasShortcode` check.

content/en/resources/keps/_index.md
```diff
@@ -0,0 +1,22 @@
+---
+linktitle: Enhancements
+title: Kubernetes Enhancement Proposals (KEPs)
+description: |
+  List of enhancements that are proposed
```
sftim, line 5:
```suggestion
   List of Kubernetes enhancements.
```
“proposed” doesn't capture that some of these are fully implemented.

palnabarun:
@sftim -- we are generating anything here. The KEPs list is pre-generated using a post submit for k/enhancements. We just consume the JSON here.

Not generating anything at build time for contributor-site reduces the complexity.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,76 @@
...
+              {{ range $index, $author := $data.authors }}
+              {{- if $index -}}, {{ end -}}
+              <a href="https://github.com/{{ trim $author "@" }}">
+                <span class="github-user">@{{ trim $author "@" }}</span>
+              </a>
+              {{ end }}
+            </td>
+            <td>
+                {{ if $data.creationDate }}
+                  {{ $data.creationDate }}
```
sftim, line 52: nit: this could be a `<span>` or a `<time>` with a [`datetime`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) attribute.

sftim:
I think this is OK.

/hold

There might be other pieces that need to land before this goes in.

- I'd prefer to be able to list all the KEPs on the final page, rather than only see 100 at a time maximum. We can fix that later.
- The sig names should be, eg “SIG Node” rather than “sig-node”. We can apply some deterministic capitalization rules to make that happen (ideally, during JSON generation: make an extra field with the right capitalization).
- For me, the links to pages like https://kep.k8s.io/1234 trigger a certificate name mismatch warning. We need a new cert with the right subjectAltName, I think.

/lgtm
(other folk: feel free to cancel that LGTM if any of those concerns justify it, or you spotted some other problem)

layouts/shortcodes/keps-data.html
palnabarun: TIL. Thanks!

content/en/resources/keps/_index.md
palnabarun: In essence, all enhancements are proposals in various states of implementation, but I agree that the language can create some confusion.

sftim:
/lgtm

Can we skip KEPS with a KEP ID < 1?

palnabarun:
> I'd prefer to be able to list all the KEPs on the final page, rather than only see 100 at a time maximum. We can fix that later.

The selector contains a "All" option now to cater to this.

> The sig names should be, eg “SIG Node” rather than “sig-node”. We can apply some deterministic capitalization rules to make that happen (ideally, during JSON generation: make an extra field with the right capitalization).

I like this idea. I would prefer keeping just the SIG name though in this form "Node" since the table header is itself SIG. Let me try something here.

> For me, the links to pages like https://kep.k8s.io/1234 trigger a certificate name mismatch warning. We need a new cert with the right subjectAltName, I think.

Moved to https://features.k8s.io/<NNNN> to resolve this for now.

palnabarun:
> Can we skip KEPS with a KEP ID < 1?

I am going to fix those. Those KEPs needs to have their metadata updated. There's also a KEP with number "NNNN".

palnabarun:
> I like this idea. I would prefer keeping just the SIG name though in this form "Node" since the table header is itself SIG. Let me try something here.

Tried a middle ground solution. `strings.TrimPrefix "sig-" $data.owningSig | humanize` would render `sig-contributor-experience` to `Contributor experience`. Ideally I would want first letter of every word to be capitalized.

sftim:
For commit 98ba82ddcfd2a4f6e106ee5191ab93e1eb248c66, please (this is important) explain _why_ you are setting a different build interval.

sftim:
Try `title` rather than `humanize` @palnabarun

.github/workflows/netlify-periodic-build.yml
```diff
@@ -1,12 +1,12 @@
 name: Trigger netlify build
 on:
   schedule:
-  - cron: '30 3 * * *'
+  - cron: '0 */6 * * *'
```
palnabarun, line 4: Reason for changing this: To improve the freshness of data on the KEPs page. We are changing the rebuild cadence from once a day to 4 times a day.

palnabarun: @sftim -- added the reasoning here.

palnabarun:
> Try title rather than humanize

Using either of them wouldn't have solved the problem since `api-machinery` would transform to `Api-Machinery` if only using title.

I solved the problem by using both in order.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,81 @@
...
+        <th scope="col" class="col-md-2">SIG</th>
+        <th scope="col" class="col-md-2">Author</th>
+        <th scope="col" class="col-md-1">Created</th>
+        <th scope="col" class="col-md-1">Last Updated</th>
+        <th scope="col" class="col-md-1">Milestone</th>
+      </tr>
+    </thead>
+    <tbody>
+      {{ $keps := getJSON "https://storage.googleapis.com/k8s-keps/keps.json" }}
+      {{ range $index, $data := $keps }}
```
PushkarJ, line 30: I think you can merge these two in one line, unless I missed something. 

Example: https://github.com/kubernetes/website/pull/35228/files#diff-dd3aae73f0f6ad9c85ca442d99c36734b43e4950f105ad5517e7f879abc02af4R9

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,81 @@
...
+                {{ printf "%s" $data.kepNumber }}
+              </a>
+            </td>
+            <td>
+              <a href="https://github.com/kubernetes/enhancements/tree/master/keps/{{$data.owningSig}}/{{$data.name}}#summary">
+                {{ printf "%s" $data.title }}
+              </a>
+            </td>
+            <td>
+              <a href=" https://github.com/kubernetes/community/blob/master/{{ $data.owningSig }}/README.md">
```
PushkarJ, line 43: Might be a good idea to make branch names and/or repo URLs `Site` variables (in `config.toml`) so when `master` -> `main` happens or if the repo is renamed, we have to switch it in one place only.

PushkarJ (commented):
This is great. Just a couple of minor suggestion (okay to skip making those changes before merging)

.github/workflows/netlify-periodic-build.yml
sftim: Could you put the reasoning _in the file_ (so it's obvious even if you don't got lack to look for a comment on this PR)?

layouts/shortcodes/keps-data.html
sftim: I'm fine with either approach.

layouts/shortcodes/keps-data.html
sftim: We could define these repo and branch names at the top of the shortcode (that's what I'd do).

content/en/resources/keps/_index.md
```diff
@@ -0,0 +1,22 @@
...
+menu:
+  main:
+    weight: 40
+---
+
+These data come from the [k/enhancements][k-enhancements]
+repository on GitHub.
+
+> Note: If you notice an empty cell in the table, the KEP may not have updated information.
+Please raise an issue in the [k/enhancements][k-enhancements] repository.
```
sftim, line 16: If possible, avoid using a block quote to mark a note. Try this instead:

```suggestion

Comment (sftim) [reply content/en/resources/keps/_index.md:16]:
```suggestion
{{% alert title="Note" color="info" %}}
If you notice an empty cell in the table, the KEP may not have updated information.
Please raise an issue in the [k/enhancements][k-enhancements] repository.
{{% /alert %}}
```

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,81 @@
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.12.1/css/jquery.dataTables.css">
```
sftim, line 1: This belongs elsewhere in the layout. Consider using [`.HasShortcode`](https://gohugo.io/templates/shortcode-templates/#checking-for-existence) to make that inclusion conditional, and leave a comment here to signpost folks to where it's done.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,81 @@
...
+<script type="text/javascript">
+  $(document).ready( function () {
+    $('#keps-table').DataTable({
+      lengthMenu: [
+        [10, 25, 50, -1],
+        [10, 25, 50, 'All'],
+      ],
+    });
+});
+</script>
```
sftim, line 13: I'd move this to the `<head>` of the page layout that gets used to render this page as HTML, and put it behind a `.HasShortcode` check.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,81 @@
...
+                  <time datetime="{{ $data.lastUpdated }}">{{ $data.lastUpdated }}</time>
+                {{ end }}
+              </span>
+            </td>
+            <td>
+              <ul class="list-unstyled">
+                {{ range $key, $value := $data.milestone }}
+                  {{ if $value }}
+                  <li>
+                    {{ $key }}:<b>{{ $value }}</b>
```
sftim, line 71: (nit):
```suggestion
                    {{ $key }}:<span class="kep-milestone-stage">{{ $value }}</span>
```

and then update the CSS to make this bold.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,81 @@
...
+            </td>
+            <td>
+                {{ if $data.lastUpdated }}
+                  <time datetime="{{ $data.lastUpdated }}">{{ $data.lastUpdated }}</time>
+                {{ end }}
+              </span>
+            </td>
+            <td>
+              <ul class="list-unstyled">
+                {{ range $key, $value := $data.milestone }}
```
sftim, line 68: Could we use more meaningful names than `key` and `value` here?

dims:
howdy folks! how far away are we from merging this and then iterating? 🙏🏾

sftim:
The one thing I'd like to see added is the comment I asked for in https://github.com/kubernetes/contributor-site/pull/324/files#r945321372

Other details we could fix post-merge. I'm asking for the comment because if that post-merge tidying never happens, we actually lose some important context.

.github/workflows/netlify-periodic-build.yml
palnabarun: Makes sense. Added a comment to the workflow file.

layouts/shortcodes/keps-data.html
palnabarun: Done

content/en/resources/keps/_index.md
palnabarun: Done.

palnabarun:
I have catered to all the suggestions except the one about the location of the stylesheet and scripts specific to the shortcode.

@sftim -- I agree that having the stylesheet and scripts in the body is an antipattern and shouldn't be done. On checking where the header is defined, it's in the theme layouts. So, moving the stylesheet and scripts there would mean replicating the `head.html` and `head-css.html` partial from the theme to the sites `layout/partials` directory. In future, this may lead to maintenance overhead. What do you suggest to mitigate this issue?

content/en/resources/keps/_index.md
palnabarun: This is a nifty TIL for me. Thanks for the tip!

jberkus:
Can you remove the top-level menu item on the home page?

content/en/resources/keps/_index.md
```diff
@@ -0,0 +1,23 @@
...
+  main:
+    weight: 40
+---
+
+These data come from the [kubernetes/enhancements]
+repository on GitHub.
+
+{{% alert title="Note" color="info" %}}
+If you notice an empty cell in the table, the KEP may not have updated information.
+Please raise an issue in the kubernetes/enhancemenets repository.
```
jberkus, line 17:
```suggestion
Please raise [an issue in the kubernetes/enhancements](https://github.com/kubernetes/enhancements/issues) repository.
```

content/en/resources/keps/_index.md
```diff
@@ -0,0 +1,23 @@
...
+  List of Kubernetes enhancements
+type: docs
+menu:
+  main:
+    weight: 40
+---
+
+These data come from the [kubernetes/enhancements]
+repository on GitHub.
+
```
jberkus, line 14:
```suggestion

To see the original Enhancement issue, click on the KEP number.  To see a summary of what the KEP does, click on the Title.

```

... this behavior is useful, but unintuitive, so adding text explaining it helps.

sftim:
> replicating the `head.html` and `head-css.html` partial from the theme to the sites `layout/partials` directory.

There are some ways in Hugo to implement code reuse. If we find there's a lot of maintenance overhead, we can consider that.

https://github.com/kubernetes/website/ uses the same Docsy theme (much more customized) and you might find some tips there about how to override just part of an inherited partial.

For scripts, you can often trigger these to load from the page _footer_.

layouts/partials/scripts.html
```diff
@@ -0,0 +1,35 @@
...
+    $(document).ready(function () {
+        $('#keps-table').DataTable({
+            lengthMenu: [
+                [10, 25, 50, -1],
+                [10, 25, 50, 'All'],
+            ],
+            order: [[2, 'asc'], [0, 'asc']],
+        });
+    });
+</script>
```
sftim, line 34: We might be able to put this into the `hooks/body-end` partial instead - @palnabarun WDYT?

(see 2 lines above for where that gets called)

sftim:
A tip: edit the PR description to include a link to the [preview](https://deploy-preview-324--kubernetes-contributor.netlify.app/resources/keps/)

layouts/partials/scripts.html
palnabarun: Both the checks now work.

palnabarun: I moved the stylesheet to `hooks/head-end.html` as well

palnabarun:
@sftim @jberkus -- this should be good to merge now?

layouts/partials/hooks/body-end.html
```diff
@@ -0,0 +1,25 @@
+{{/*
+    This file is a 1:1 copy of the one in docsy but with adding scripts
```
sftim, line 2: nit: which Docsy release / commit is this from?

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,76 @@
...
+  and `partials/hooks/body-end.html` respectively.
+
+  Note: In order to configure Datatable parameters for changing any behavior, edit
+  the `DataTable` initialization in `partials/hooks/body-end.html`
+
+*/}}
+
+<div>
+  <table id="keps-table" class="table table-hover mt-2 col-md-12 table-sm">
+    <thead class="thead-light">
```
sftim, line 13: Optionally: add a table caption.

layouts/shortcodes/keps-data.html
```diff
@@ -0,0 +1,76 @@
...
+  The stylesheets and scripts required are in `partials/hooks/head-end.html`
+  and `partials/hooks/body-end.html` respectively.
+
+  Note: In order to configure Datatable parameters for changing any behavior, edit
+  the `DataTable` initialization in `partials/hooks/body-end.html`
+
+*/}}
+
+<div>
+  <table id="keps-table" class="table table-hover mt-2 col-md-12 table-sm">
```
sftim, line 12: nit: If someone were to use that shortcode twice in one page, the output would have a duplicated `id`. I think this is low risk low impact.

sftim (commented):
/lgtm

mrbobbytables (commented):
/approve

sftim:
Let's ship it

/hold cancel

PR #333, opened 2022-08-27
Adding late-breaking change about testing option

bridgetkromhout:
I noticed when I got the email about the contributor summit that it explained the late-breaking changes to the health and safety policy, but then it linked me to a registration page which only listed the previous policy.

I want to ensure that when people register for the contributor summit, they don't have a false sense of security caused by assuming that the summit has stricter requirements than the conference itself, so I am submitting a PR to clarify the change.

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>


bridgetkromhout:
Note that this discrepancy is also present in the cvent pages during the registration flow:

image

jberkus:
@bridgetkromhout reported to CNCF

PR #343, opened 2022-10-11
[KCSNA2022] CFP is no longer open

detiber:
Update the Schedule page for KCSNA2022 to remove statement that the CFP is open.


mrbobbytables (commented):
/lgtm
/approve

PR #353, opened 2022-11-02
chore: add `.dockerignore`

Pradumnasaraf:
#### What this PR does / why we need it:

- This PR will add a `.dockerignore` file. We need it to reduce the docker image build size and make the build faster. Also good for local development


jberkus:
Looks good, thanks!

/lgtm
/approve

Issue #93, opened 2020-01-19
Update method in which external content is ingested.


mrbobbytables:
Currently, external content is brought into the contributor-site via the [hack/gen-content.sh]() bash script. In its original inception, it was quite small and lean -- but as a bash script usually does, it has grown to be unwieldy. 

It'd be far better to replace it with a more supportable (and testable) means of bringing in external content. This means rewriting it something like go or python, or finding some other means of bringing in external content.

/kind feature
cc @vishakhanihore @nimbinatus

markyjackson-taulia:
this is a long term replacement and is not needed for the contributor site.

nimbinatus:
Hey @markyjackson-taulia! I had started poking at this issue a couple weeks ago to see what we needed. I'm pretty sure a simple include will work with a bit of tweaking. Did you want to see what I had in draft form?

markyjackson-taulia:
@nimbinatus sorry for the delay in responding. I would be interested to see what you have. I am thinking of converting this in its entirety to go so I can include some testing around it.

nimbinatus:
@markyjackson-taulia sounds good! I'll try to clean up my code with some comments and get my branch up this week for you to poke at.

nimbinatus:
@markyjackson-taulia here's what I was poking at. It does require a bit more work to add pages to the repo, and it's not 100% accurate yet regarding some of the internal links on each page. However, it's more native to Hugo, so I was hoping it would be a bit cleaner to test.

https://github.com/nimbinatus/contributor-site/blob/shortcode-import/layouts/shortcodes/content.gohtml

markyjackson-taulia:
Will run some test

mrbobbytables:
The hardest part that I'm concerned about with the hugo method is the replication of the link expander. It determines if the content is has been synced locally and updates it with the correct path, or if not -- expand it to the full github url.

This blurb is slightly out of date (updated in #110), but it is similar enough to explain.
https://github.com/kubernetes-sigs/contributor-site/blob/fe36c430d92ed0a81cb85e154fa092f55912a3be/hack/gen-content.sh#L68-L89

markyjackson-taulia:
I found that through my testing, using the hugo method did not do fully what I wanted so I am going to redo this. Thank you @nimbinatus for the code

markyjackson-taulia:
@nimbinatus I started the work in the above PR, did it in python but there is still more to do. 
I would love to pair with you on this.

[...]

PR #366, opened 2023-01-11
Allow leading spaces for reference links

stmcginnis:
Some of the sites that get pulled in use a mix of inline and reference links. Reference links are expected to have the format of the link reference name in brackets, starting at the beginning of the line, followed by a colon and then the linked target.

Some of the external sites have been found to have some variation on this format though. While markdown references show the format as above, it does work when these link references have leading spacing before the opening bracket. This has caused [broken slack links](https://github.com/kubernetes/community/issues/6934), as one example.

This change updates the regex for identifying these reference links in the get-content script to allow for leading spaces.

### DIFF
--- a/hack/gen-content.sh
+++ b/hack/gen-content.sh
@@ -115,7 +115,7 @@ process_content() {
   fi
 
   mapfile -t ref_link_matches < \
-    <(grep -o -i -P '^\[.+\]:\s*(?!mailto|\S+?@|<|>|\?|\!|@|#|\$|%|\^|&|\*)\K\S+$' "$1")
+    <(grep -o -i -P '^\s+\[.+\]:\s*(?!mailto|\S+?@|<|>|\?|\!|@|#|\$|%|\^|&|\*)\K\S+$' "$1")
 
   if [[ -v ref_link_matches ]]; then
     for match in "${ref_link_matches[@]}"; do


mrbobbytables (commented):
Hm..while we don't have a tests for it - I did a manual check and it seems to break other links =/ 

here is live: https://www.kubernetes.dev/docs/guide/#community-expectations-and-roles
here is a preview: https://deploy-preview-366--kubernetes-contributor.netlify.app/docs/guide/#community-expectations-and-roles

The 'fuzzyness' of markdown and the regexes have been the bane of the ingestion script -_-

There have been a few attempts to rewrite everything in go and use a native markdown parser, but those have also had other issues.

stmcginnis:
Hmm... there are definitely parts where this is not correct. I can see some cases where there are example markdown comments in docstrings that could be problematic here. There's likely other conditions.

It looks like maybe the safest thing for now would be to just enforce that legitimate link references do not have leading characters. I can try to watch out for those and help catch any incoming issues in reviews.

In the meantime, I'll watch #93 and see if there is any progress there. If I have a good time block to focus on it, I may take that over and give it one more shot. Currently assigned, but that was quite awhile ago and there hasn't been any progress. So likely they didn't realize the amount of work they were signing up for with that one. :D 

Will close this for now since it definitely is the best approach. Thanks for taking a look!

Issue #360, opened 2022-11-11
Upgrade Hugo and Docsy


chalin:
The versions of Hugo and Docsy are quite old.

https://github.com/kubernetes/contributor-site/blob/1a8d9c6dff430129d1ab6998b539f8367aa9171e/netlify.toml#L5-L6

For example, Hugo [0.69.2](https://github.com/gohugoio/hugo/releases/tag/v0.69.2) was released in April 2020.

/cc @nate-double-u

chalin:
Why upgrade? E.g., Docsy's current page-feedback feature is incompatible with GA4 (https://github.com/google/docsy/issues/1302). Since UA is being deprecated and this site is being migrated to GA4 (#354), the a new version of Docsy will eventually be needed. (There is a work-around for the page-feedback issue, so the upgrade isn't urgent.)

jberkus:
/remove-lifecycle stale

We were able to upgrade to 0.89, but remain stalled on the upgrade to 110.

sftim:
See the revised PR, #374

chris-short:
We did spend some time on this during the Contributor Summit at KubeCon Chicago. [@jeefy's fork](https://github.com/jeefy/contributor-site/tree/jeefy-update-hugo) has some errors we need to work through but, we've made a step forward here.

sftim:
What still needs to happen before we can call this Done?

jeefy:
I'm back and working through said errors (actually I started from scratch and scrapped my post-lunch hastily-done branch)

Most errors that we encountered at the summit are fixed, and the last thing I'm updating is the Dockerfile. PR inc soon.

jeefy:
Oh duh.

/assign

chris-short:
Were you going to tackle the move from submodules to Hugo Modules? I can do that separately if you'd like.

I also want to update the K8s.dev picture now that we have a big crowd pic for a replacement.

jeefy:
Kinda did it all, see here: https://github.com/kubernetes/contributor-site/pull/456 

Need help with a CSS issue post-migration but otherwise it's all done and clean.

stmcginnis:
Looks like we're up to [Hugo 0.124 now](https://github.com/kubernetes/contributor-site/blob/c9d1a78cbed3609a0db62f410d7bc33d36cbce07/netlify.toml#L6). There is a slightly newer release out, but I think it's just regular dependency management at this point. Follow-on work being done in #456. I think it's safe to close this issue now.

/close

PR #374, opened 2023-02-18
Bump Hugo version to 0.112

sftim:
Bump Hugo version to ~0.110~ 0.112 ([preview](https://deploy-preview-374--kubernetes-contributor.netlify.app/)).

There is build-time warning:

<pre>
<i><b>.Path</b> when the page is backed by a file is deprecated and will be removed in a future release. We plan to use <b>Path</b> for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct similar to the one below:</i>

  {{ $path := "" }}
  {{ with .File }}
	{{ $path = .Path }}
  {{ else }}
	{{ $path = .Path }}
  {{ end }}
</pre>

Also, update local preview using a container image (copy approach from [k/website](https://github.com/kubernetes/website) `Makefile`). The container image now mounts this Git repo read-only.

Helps with https://github.com/kubernetes/contributor-site/issues/360

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ RUN apk add --no-cache \
     curl \
     git \
     grep \
+    gcompat \
     libc6-compat \
     rsync \
     sed \
--- a/Makefile
+++ b/Makefile
@@ -75,12 +75,17 @@ docker-server:
 	$(MAKE) container-server
 
 container-server: ## Run Hugo locally within a container, available at http://localhost:1313/
+	# no build lock to allow for read-only mounts
 	git submodule update --init --recursive --depth 1
 	$(CONTAINER_RUN) -p 1313:1313 \
 		--mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 \
+		--read-only \
+		--cap-drop=ALL \
+		--cap-drop=AUDIT_WRITE \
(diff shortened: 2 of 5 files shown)


mrbobbytables:
/hold
This does break the site a bit - take a look at the side menu on the preview vs the actual site:
https://deploy-preview-374--kubernetes-contributor.netlify.app/docs/
https://kubernetes.dev/docs/

jberkus:
@sftim so the issue is the preview rather than the Hugo version?

sftim:
I think there is a real problem here; it's harder to see what that is because my local preview is - right now - missing the generated content either way.

We shouldn't merge this change as-is. 

PR https://github.com/kubernetes/contributor-site/pull/375 may help us to find a newer Hugo version that nonetheless works OK, and that find may then help us spot what to change so that we can upgrade further.

sftim:
We can work around the Hugo problem by setting `GOMAXPROCS` to 1

sftim:
I think this is good to unhold - check the site, it looks right with the latest fixes.

We can remove those once a separate PR updates Docsy.

sftim:
Thanks

/hold cancel

fsmunoz:
FWIW I've compared things side-by-side and didn't find any different, looks good.

PR #375, opened 2023-02-19
Generate content even for container preview

sftim:
This change makes sure that generated content is up to date when previewing locally in a container.


sftim:
I'm going to use this as a basis for my work on #374 (whether we merge it or not).

hack/gen-content.sh
sftim, line 1: I've changed this script so that - at least, in theory - it works equally well for a local build as for a container preview.

sftim:
:thought_balloon: do we always want this behavior, or (eg) only when a particular variable is set?

jberkus:
What's the status of this PR?  It doesn't sound like you're ready with it.

sftim:
This works for me @jberkus

I don't know whether it's useful for other contributors, and I can't easily answer that question.

I'll rebase it.

Issue #355, opened 2022-11-11
Google Analytics 4 migration followup


chalin:
This is a followup item to

- #354

### Tasks

- [ ] Drop UA ID once it is no longer receiving events (in 2H2023) or are no longer needed. Given that this site uses the Docsy page-feedback feature, you'll need to add a bogus ID to the config file (at least, that is the required behavior as of the time of writing this).
- [x] #356

PR #376, opened 2023-02-23
Drop Google Universal Analytics (UA) ID

chalin:
- Contributes to #355
- Ideally, we'd drop the `services.googleAnalytics` config parameter entirely, but that would require an update of the layouts that refer to it. I'll leave that cleanup for later.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -31,7 +31,8 @@ anchor = "smart"
 
 [services]
 [services.googleAnalytics]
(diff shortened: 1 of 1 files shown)


jberkus:
Just checking, this works even without the Hugo upgrade?

jberkus (approved):

jberkus:
/approve

need an LGTM from another contributor ... @jeefy ?

PR #381, opened 2023-03-08
Add blog post for scheduler plugin reusing

uniemimu:


jberkus:
Apolgies!  Somehow this PR never showed up in my notifications, and as such I just noticed it.  Thank you so much for your patience.  I will review it during/after Kubecon and move it forwards.

jberkus:
This passes spelling and composition check, but I'm waiting for @kubernetes/sig-scheduling-misc review; I've reached out to the SIG.

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,124 @@
...
+
+Resource calculations and scoring of nodes is currently done quite nicely and with a well configurable interface within the _NodeResourcesFit_ plugin. It can be [configured](https://kubernetes.io/docs/reference/config-api/kube-scheduler-config.v1/#kubescheduler-config-k8s-io-v1-NodeResourcesFitArgs) to use different scoring strategies such as _LeastAllocated_ and _MostAllocated_. It does the job, and it does it well. Reinventing the wheel which has been implemented in _NodeResourcesFit_ for calculating of node resources, would seem like a lot of unnecessary, duplicated work. Ideally, one would take up topology resource calculations and scoring from where _NodeResourcesFit_ plugin left off, and form the topology score based on the scores already calculated for individual nodes.
+
+## Alternative approaches for plugin reuse
+
+### Ctrl-C, Ctrl-V
+
+> **Warning**
+> The simplest and probably most frowned upon code-reuse practice is of course copy-paste. Due to well known issues with copy-paste code, this approach can't be recommended, and is mentioned here only as an example of what not to do. Copy-pasting won't be any faster anyway, as you will see below.
```
Huang-Wei, line 21: maybe remote the `>` symbol? otherwise this section is entirely quoted without any regular paragraph, which looks a bit strange.

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,124 @@
...
+* Could multiple new plugins utilize scores from the same plugin, without running the same scoring logic N times?
+* Could the dependency to the reused plugin be made less fragile, reducing the coupling?
+* Could reuse be achieved without duplicated args and configuring same algorithms the same way, twice (or more)?
+* Could things be configured more dynamically instead of hard-coding to the sources one plugin uses another plugin?
+
+These questions had the author thinking about finding a better solution than just composition. Also the documentation of the [framework.CycleState](https://github.com/kubernetes/kubernetes/blob/2f977fd8c400971deed365dcc437e519bce475f0/pkg/scheduler/framework/cycle_state.go#L42-L47) was looking rather tempting.
+
+> **Warning**
+> The following section is speculative and based on proof-of-concepting by the author. It doesn't represent what you can do with the Kubernetes scheduler at the time of writing. Some day the concept below might be part of Kubernetes, should the PoC develop into a successful KEP. For now, this is merely a scheduling-sig discussion item which shows the potential of the CycleState.
```
Huang-Wei, line 84: scheduling-sig -> sig-scheduling

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,124 @@
...
+
+Ideally, a new plugin could ask for getting the scores from another plugin, or multiple plugins, without rerunning them. The tricky thing is, that plugins run highly parallel in order to execute faster. Thus proper synchronization would be needed. Also it would be ideal, if no changes to existing plugins would be required, and the framework would provide for plugins in need of reusing other plugins' results.
+
+This got the author thinking, that if the `framework.CycleState` object could contain a channel for the scores, and if the framework would create such a channel when it is needed, then the parallel running plugins which are interested in score reuse, could wait for the scores to appear from such "ScoreChannels".
```
Huang-Wei, line 88: I'm trying to understanding why we need a channel. If we place the composite plugin in the end (or at least place after their dependent plugins), and then previous plugins write their {pluginName: score} to the CycleState, then I'd assume the composite plugin can reuse those pre-calculated scores already?

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,124 @@
...
+	C chan framework.NodeScoreList
+}
+```
+
+A little less than 80 framework lines of PoC-code later the now very dirty scheduler was able to pick up from the arguments of plugins, whether they wanted to use scores from other plugins, and it was creating such "ScoreChannels" into the `framework.CycleState` and wrote scores to them as requested. A couple of experimental plugins later it was possible to verify that indeed, plugin scoring graphs such as the following were entirely possible:
+
+{{< figure src="diagram1.svg" alt="Plugins A, B, C and D with connections A->B, A->C, B->D, C->D">}}
+
+Which makes a nice demo of the one-to-many and many-to-one pipelining possibilities for plugin score reusing.
+
```
Huang-Wei, line 102: could you put a pseudo code snippet here to explain how TopologyScorer leverages the score it get?

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
uniemimu: The channel is only needed for concurrency, and it isn't really necessary at all if one does a composite plugin which reruns the logic of the other plugin. If you rerun the whole other plugin in a composite plugin, you get its score that way, then there's no need to fetch anything from cycle state. The code is reused by running it again, but there is no result reuse.

The point of the channel usage is to not do a composite, and instead run the source plugin only once as it is, and then reuse its result score later. As mentioned in the quoted paragraphs, the score plugins run parallel, and to be able to reuse scores from another plugin, one needs to synchronize. Using a channel is the easy way. The plugin which is in the receiving end waits for the score to appear from the channel. That appearance ensures that the reused plugin has completed processing the score. One could perhaps alternatively poll something in the cycle state object, but that wouldn't be tidy.

You could of course have dedicated processing steps, like "postScore" which runs after "score" and "postPostScore" which runs after "postScore", and then you'd be able to reuse the results from prior steps which could be stored in the cycle state, but why do a bigger framework change like that, if one can do a simple channel solution which allows for a generic solution (unlimited chain).

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
uniemimu: Ok, will do.

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
uniemimu: sure

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
uniemimu: Ok, will do.

jberkus:
OK, this looks good, should we publish?

Huang-Wei:
> OK, this looks good, should we publish?

Yup,
/lgtm on scheduling side.

uniemimu:
Should I squash this?

jberkus:
You can if you want.  I can also have Tide do it.

However, I need to confer with @sftim  on how we fix the date for the blog post.

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,161 @@
...
+> **Warning**
+> The following section is speculative and based on proof-of-concepting by the author. It doesn't represent what you can do with the Kubernetes scheduler at the time of writing. Some day the concept below might be part of Kubernetes, should the PoC develop into a successful KEP. For now, this is merely a sig-scheduling discussion item which shows the potential of the CycleState.
+
+Ideally, a new plugin could ask for getting the scores from another plugin, or multiple plugins, without rerunning them. The tricky thing is, that plugins run highly parallel in order to execute faster. Thus proper synchronization would be needed. Also it would be ideal, if no changes to existing plugins would be required, and the framework would provide for plugins in need of reusing other plugins' results.
+
+This got the author thinking, that if the `framework.CycleState` object could contain a channel for the scores, and if the framework would create such a channel when it is needed, then the parallel running plugins which are interested in score reuse, could wait for the scores to appear from such "ScoreChannels".
```
sftim, line 87:
```suggestion
This got me thinking that if the `framework.CycleState` object could contain a channel for the scores, and if the framework would create such a channel when it is needed, then the parallel running plugins which are interested in score reuse could wait for the scores to appear from such `ScoreChannels`.
```

even better rewritten as shorter sentences (commas are not enough; a rewrite would use more periods)

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,161 @@
...
+```go
+// ScoreChannel for giving score results to other plugins.
+type ScoreChannel struct {
+	C chan framework.NodeScoreList
+}
+```
+
+A little less than 80 framework lines of PoC-code later the now very dirty scheduler was able to pick up from the arguments of plugins, whether they wanted to use scores from other plugins, and it was creating such "ScoreChannels" into the `framework.CycleState` and wrote scores to them as requested. A couple of experimental plugins later it was possible to verify that indeed, plugin scoring graphs such as the following were entirely possible:
+
+{{< figure src="diagram1.svg" alt="Plugins A, B, C and D with connections A->B, A->C, B->D, C->D">}}
```
sftim, line 98:
```suggestion
{{< figure src="diagram1.svg" alt="Plugins A, B, C and D with connections A → B, A  → C, B  → D, C  → D. The connections are via a Cycle State element." >}}
```

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Summary
+
+There are several ways to do Kubernetes scheduler plugin reusing, ranging from strictly not recommended copy-pasting to composition and all the way to modifying the scheduler framework, depending on how efficient you need things to be. The scheduler framework includes built-in utilities such as the `CycleState`, and by no means does it prevent plugins from passing information between each other. The pieces are there, only thing currently missing is perhaps a bit of assisting from the side of the Scheduler framework runtime.
+
+## Next steps
+
+As per discussed in the sig-scheduling meetings, the topology resource scoring plugin will proceed with the more conservative composition approach, because it can get the job done without framework changes, associated risks and maintenance burden. However, if you find that it would indeed benefit your use-cases to be able to simply configure scheduler profiles in such a way that certain score plugins are connected, chime in at slack [#sig-scheduling].
```
sftim, line 159:
```suggestion
As per discussed in SIG Scheduling meetings, the topology resource scoring plugin will proceed with the more conservative composition approach, because it can get the job done without framework changes, associated risks and maintenance burden. However, if you find that it would indeed benefit your use-cases to be able to simply configure scheduler profiles in such a way that certain score plugins are connected, chime in via [`#sig-scheduling`][#sig-scheduling] on Kubernetes Slack (visit https://slack.k8s.io/ for an invitation).
```

content/en/blog/2023/2023-03-08-scheduler-plugin-reusing/index.md
```diff
@@ -0,0 +1,161 @@
...
+This got the author thinking, that if the `framework.CycleState` object could contain a channel for the scores, and if the framework would create such a channel when it is needed, then the parallel running plugins which are interested in score reuse, could wait for the scores to appear from such "ScoreChannels".
+
+```go
+// ScoreChannel for giving score results to other plugins.
+type ScoreChannel struct {
+	C chan framework.NodeScoreList
+}
+```
+
+A little less than 80 framework lines of PoC-code later the now very dirty scheduler was able to pick up from the arguments of plugins, whether they wanted to use scores from other plugins, and it was creating such "ScoreChannels" into the `framework.CycleState` and wrote scores to them as requested. A couple of experimental plugins later it was possible to verify that indeed, plugin scoring graphs such as the following were entirely possible:
```
sftim, line 96:
```suggestion
A little less than 80 framework lines of PoC-code later the now very dirty scheduler was able to pick up from the arguments of plugins, whether they wanted to use scores from other plugins, and it was creating such `ScoreChannels` into the `framework.CycleState` and wrote scores to them as requested. A couple of experimental plugins later it was possible to verify that indeed, plugin scoring graphs such as the following were entirely possible:
```

uniemimu:
> @uniemimu do you have capacity to address Tim's feedback?

Obviously I haven't had, my apologies for that. Since the somewhat related KEP PR (3925) never found any interest, I'm leaning towards closing this PR without publishing. What do you think?

natalisucks:
@uniemimu I think that's completely fine. If the KEP picks up traction again, this PR can always be re-opened and worked on once again. Thanks for answering!

PR #382, opened 2023-03-13
Add Mastodon Verification

chris-short:
Contributor Comms has created a K8sContributors Mastodon account on Hachyderm.io. We would like to prevent copycats and make this the "one true" K8sContributors account on Mastodon and adding this link is the verification process for Mastodon.

### DIFF
--- a/content/en/welcome/_index.md
+++ b/content/en/welcome/_index.md
@@ -72,5 +72,6 @@ There are many ways you can stay involved with the Kubernetes Contributor commun
 
 {{% /blocks/section %}}
 
-
+# Mastodon Account Verification Code
+<a rel="me" href="https://hachyderm.io/@K8sContributors"></a>
 


content/en/welcome/_index.md
```diff
@@ -72,5 +72,6 @@ There are many ways you can stay involved with the Kubernetes Contributor commun
 
 {{% /blocks/section %}}
 
-
+# Mastodon Account Verification Code
+<a rel="me" href="https://hachyderm.io/@K8sContributors"></a>
```
mrbobbytables, line 76: This is the wrong page for verification
It should go here:
https://github.com/kubernetes/contributor-site/blob/master/content/en/_index.md

It also does not need a header, the link just has to be present on the site for it to detect

chris-short: I apparently forgot how to Markdown.

PR #392, opened 2023-04-05
blog: updated E2E best practices

pohly:
"Writing good E2E tests" was already updated a while ago in https://github.com/kubernetes/community/pull/7021. As suggested there (https://github.com/kubernetes/community/pull/7021#issuecomment-1374333094), we should bring this update to the attention of more contributors, hence this blog post.

/cc @aojea @jberkus

### DIFF
--- a/content/en/blog/2023/e2e-testing-best-practices.md
+++ b/content/en/blog/2023/e2e-testing-best-practices.md
@@ -173,7 +173,7 @@ before the test ran. [`ginkgo.DeferCleanup`
 this because it can be called similar to `defer` directly after setting up
 something. It is better than `defer` because Ginkgo will show additional
 details about which cleanup code is running and (if possible) handle timeouts
-for that code (see next section). Is is better than `ginkgo.AfterEach` because
+for that code (see next section). It is better than `ginkgo.AfterEach` because
 it is not necessary to define additional variables and because
 `ginkgo.DeferCleanup` executes code in the more useful last-in-first-out order,
 i.e. things that get set up first get removed last.
@@ -281,7 +281,13 @@ When waiting for something to happen, use a reasonable timeout. Without it, a
 test might keep running until the entire test suite gets killed by the
 CI. Beware that the CI under load may take a lot longer to complete some
 operation compared to running the same test locally. On the other hand, a too
-long timeout can be annoying when trying to debug tests locally.
+long timeout also has drawbacks:
+- When a feature is broken so that the expected state doesn't get reached, a test
+  waiting for that state first needs to time out before the test fails.
+- If a state is expected to be reached within a certain time frame, then a
+  timeout that is much higher will cause test runs to be considered successful
+  although the feature was too slow. A dedicated performance test in a well-know
+  environment may be a better solution for testing such performance expectations.
 
 The framework provides some [common
 timeouts](https://github.com/kubernetes/kubernetes/blob/eba98af1d8b19b120e39f3/test/e2e/framework/timeouts.go#L44-L109)
@@ -292,9 +298,10 @@ timeout in the test.
 
 Good code that waits for something to happen meets the following criteria:
 - accepts a context for test timeouts
-- informative during interactive use (i.e. intermediate reports, either
-  periodically or on demand)
-- little to no output during a CI run except when it fails
+- depending on how the test suite was invoked:
+  - informative during interactive use (i.e. intermediate reports, either
+    periodically or on demand)
+  - little to no output during a CI run except when it fails
 - full explanation when it fails: when it observes some state and then
   encounters errors reading the state, then dumping both the latest
   observed state and the latest error is useful
--- a/content/en/blog/2023/e2e-testing-best-practices.md
+++ b/content/en/blog/2023/e2e-testing-best-practices.md
@@ -1,0 +1,407 @@
+---
+layout: blog
+title: "E2E Testing Best Practices, Reloaded"
+date: 2023-04-12
+slug: e2e-testing-best-practices-reloaded
+author: "Patrick Ohly (Intel)"
+---
+
+End-to-end (E2E) testing in Kubernetes is how the project validates
+functionality with real clusters. Contributors sooner or later encounter it
+when asked to write E2E tests for new features or to help with debugging test
+failures. Cluster admins or vendors might run the conformance tests, a subset
+of all tests in the [E2E test
+suite](https://github.com/kubernetes/kubernetes/tree/v1.27.0-rc.0/test/e2e).
+
+The underlying [E2E
+framework](https://github.com/kubernetes/kubernetes/tree/v1.27.0-rc.0/test/e2e/framework)
+for writing these E2E tests has been around for a long
+time. Functionality was added to it as needed, leading to code that became hard
+to maintain and use. The [testing commons
+WG](https://github.com/kubernetes/community/blob/master/sig-testing/README.md#testing-commons)
+started cleaning it up, but dissolved before completely achieving their
+goals.
+
+After the [migration to Gingko
+v2](https://github.com/kubernetes/kubernetes/pull/109111) in Kubernetes 1.25, I
+picked up several of the loose ends and started untangling them. This blog post
+is a summary of those changes. Some of this content is also found in the
+Kubernetes contributor document about [writing good E2E
+tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/writing-good-e2e-tests.md)
+and gets reproduced here to raise awareness that the document has been updated.
+
+## Overall architecture
+
+At the moment, the framework is used in-tree for testing against a cluster
+(`test/e2e`), testing kubeadm (`test/e2e_kubeadm`) and kubelet
+(`test/e2e_node`). The goal is to make the core `test/e2e/framework` a package
+that has no dependencies on internal code and that can be used in different E2E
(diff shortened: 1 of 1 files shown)


xmcqueen (approved):
excellent
/lgtm

pohly:
/assign @mrbobbytables 

For approval. Let's get this published before KubeCon, then folks can chat with me about it there.

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Recovering from test failures
+
+All tests should ensure that a cluster is restored to the state that it was in
+before the test ran. [`ginkgo.DeferCleanup`
+](https://pkg.go.dev/github.com/onsi/ginkgo/v2#DeferCleanup) is recommended for
+this because it can be called similar to `defer` directly after setting up
+something. It is better than `defer` because Ginkgo will show additional
+details about which cleanup code is running and (if possible) handle timeouts
+for that code (see next section). Is is better than `ginkgo.AfterEach` because
```
aojea, line 176:
```suggestion
for that code (see next section). It is better than `ginkgo.AfterEach` because
```

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+it is not necessary to define additional variables and because
+`ginkgo.DeferCleanup` executes code in the more useful last-in-first-out order,
+i.e. things that get set up first get removed last.
+
+Objects created in the test namespace do not need to be deleted because
```
aojea, line 181: maybe you should introduce this at the beginning of the section, that the framework creates a test namespace to avoid test pollution ... I don't know if this behavior of "test namespaces" is well known outside of kubernetes

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+deleting the namespace will also delete them. However, if deleting an object
+may fail, then explicitly cleaning it up is better because then failures or
```
aojea, line 183: also, if you create objects that are going to take some times to be completed delated, I had this mistakes with "terminating pods", I set a grace period of 300 seconds and the pod blocks the garbage collectors for that time

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+timeouts related to it will be more obvious.
+
+In cases where the test may have removed the object, `framework.IgnoreNotFound`
```
aojea, line 186: indeed this is a common mistake

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+
+        ginkgo.It("something", ginkgo.NodeTimeout(30 * time.Second), func(ctx context.Context) {
+        })
+
+  There is also a `ginkgo.SpecTimeout`, but that then also includes the time
+  taken for `BeforeEach`, `AfterEach` and `DeferCleanup` callbacks.
+
+## Polling and timeouts
+
+When waiting for something to happen, use a reasonable timeout. Without it, a
```
aojea, line 280:
```suggestion
When waiting for something to happen and you need to do asynchronous assertions, use a reasonable timeout. Without it, a
```

aojea: o asynchronous checks, I think people is familiar with this term

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+test might keep running until the entire test suite gets killed by the
+CI. Beware that the CI under load may take a lot longer to complete some
+operation compared to running the same test locally. On the other hand, a too
```
aojea, line 283: " On the other hand,"  here is misleading, I think that you may express the too main problems
- short timeout, the test will flake, per example if the CI is slow
- long timeout, the test may hide underline issues, per example, if there are some races with other components and eventually the condition pass

The thing with timeouts, is that you also should define what is the expected time you consider valid for a an operation to succeed, e2e are not only functiona,  i.e. creating a pod and it takes more than 10 minutes to run should not pass because that environment is too busy, or Services can not take more than 1 minute in program the dataplance, ... timeouts are also important to set the upper limits for some behaviors

aojea: the key is to obtain the right balance, a timeout that doesn't flake and that the time is considerable acceptable for that operation

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+- informative during interactive use (i.e. intermediate reports, either
+  periodically or on demand)
+- little to no output during a CI run except when it fails
+- full explanation when it fails: when it observes some state and then
+  encounters errors reading the state, then dumping both the latest
+  observed state and the latest error is useful
+- extension mechanism for writing custom checks
+- early abort when condition cannot be reached anymore
+
+[`gomega.Eventually`](https://pkg.go.dev/github.com/onsi/gomega#Eventually)
```
aojea, line 304: it also accepts contexts :)

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+
+  Some of the E2E framework sub-packages have helper functions that wait for
+  certain domain-specific conditions. Currently most of these functions don't
+  follow best practices (not using gomega.Eventually, error messages not very
+  informative). [Work is
+  planned](https://github.com/kubernetes/kubernetes/issues/106575) in that
+  area, so beware that these APIs may
+  change at some point.
+
+- Use `gomega.Consistently` to ensure that some condition is true
```
aojea, line 369: this is important, unfortunately I'm afraid we are not doing it as much as we should

aojea:
LGTM
overall good summary of the work done and the expectations for e2e tests, it will be good to promote this blog for reviewers of the different SIGs , so we can have consistency on all the e2e tests.

I left some comments that are not blockers and if necessary can be follow ups

/assign @sftim 

you need docs people IIRC for approval

pohly:
/hold

I'll follow up on some of the suggestions before this gets merged.

content/en/blog/2023/e2e-testing-best-practices.md
pohly: There are two problems with too long timeouts:
- a feature is broken and some expected state will never occur, but the test needs to run till it times out waiting for that state
- a feature is normally supposed to work within a certain time frame and for some reason is taking too long

The problem is that we don't have good enough control over the performance of the clusters that we test against, nor do many features have any solid "must work within XYZ seconds". Solving that problem goes beyond what we can solve right now.

I'll clarify the first point and add a comment about the second.

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,407 @@
...
+through the [framework
+instance](https://github.com/kubernetes/kubernetes/blob/1e84987baccbccf929eba98af1d8b19b120e39f3/test/e2e/framework/framework.go#L122-L123).
+When writing a test, check whether one of those fits before defining a custom
+timeout in the test.
+
+Good code that waits for something to happen meets the following criteria:
+- accepts a context for test timeouts
+- informative during interactive use (i.e. intermediate reports, either
+  periodically or on demand)
+- little to no output during a CI run except when it fails
```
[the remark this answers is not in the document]
pohly: I'll explain that the amount of information should depend on how the E2E suite was invoked.

content/en/blog/2023/e2e-testing-best-practices.md
pohly: Which is part of "all criteria", right? So no need to change anything in the text.

pohly:
/hold cancel

PR updated, ready for another LGTM and approval.

content/en/blog/2023/e2e-testing-best-practices.md
```diff
@@ -0,0 +1,414 @@
+---
+layout: blog
+title: "E2E Testing Best Practices, Reloaded"
+date: 2023-04-04
```
sftim, line 4: For a publication date, how about 2023-04-12?

sftim (commented):
/lgtm cancel

The publication date should be in the future at the time we merge it. However, it looks otherwise OK.

sftim:
BTW, approval is SIG ContribEx blog team.

content/en/blog/2023/e2e-testing-best-practices.md
aojea: yeah, I was just a personal comment to show my +1 to this function

aojea:
/lgtm

Thanks

content/en/blog/2023/e2e-testing-best-practices.md
pohly: Works for me, updated accordingly: https://github.com/kubernetes/contributor-site/compare/846fba3cd23f37924ea22e687ad015d838a4a7fd..7f1a7e9808047ed5c84b577428ccdfab209c4d4e

mrbobbytables (commented):
/lgtm
/approve

PR #395, opened 2023-04-12
Centralizing aliases under _redirects

chris-short:
I was asked to add some redirects for the Contributor Summit. After some further investigation I found that we had two different /summit aliases. To prevent this in the future, centralizing aliases into one file is in line with the k8s.io site. Plus, it's a one stop shop for new aliases/redirects when they need to be added in the future.


mrbobbytables (commented):
/lgtm
/approve

PR #416, opened 2023-08-10
Update CNCF Community Code of Conduct v1.3

ydFu:
The current upstream CNCF Code of Conduct has been updated to version 1.3.
Therefore, in this pull request, we have incorporated the updates based on [CNCF Community Code of Conduct v1.3](https://github.com/cncf/foundation/blob/c79711b5127e2d963107bc1be4a41975c8791acc/code-of-conduct.md).

Related PR: https://github.com/kubernetes/website/pull/42398

PR #421, opened 2023-08-20
Import and use the CNCF Code of Conduct

cblecker:
This uses a generated import of the CNCF CoC

Fixes #416

### DIFF
--- a/content/en/includes/_index.md
+++ b/content/en/includes/_index.md
@@ -1,0 +1,3 @@
+---
+headless: true
+---
--- a/content/en/includes/_index.md
+++ b/content/en/includes/_index.md
@@ -1,3 +1,6 @@
 ---
 headless: true
+_build:
+  list: never
+  render: false
 ---
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 CONTAINER_ENGINE	?= docker
-CONTAINER_RUN		:= $(CONTAINER_ENGINE) run --rm -it -v $(CURDIR):/src
+CONTAINER_RUN		:= $(CONTAINER_ENGINE) run --user : --rm -it -v $(CURDIR):/src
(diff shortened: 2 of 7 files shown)


sftim:
Approach seems good to me.

sftim:
I'd expect to see 2 or more commits, one of which actually does the import.

sftim:
Want to put a Markdown file in source control but not have that render to its own web page? See https://github.com/kubernetes/website/pull/41793/files#diff-6837431359d119c18686bd0377481bf69921b81f4a221801d31e7fbe57aaa642 for an example of a way to do that:

```yaml
_build:
  list: never
  render: false
```

cblecker:
@sftim This should be ready for review. Thanks!

mrbobbytables:
> Want to put a Markdown file in source control but not have that render to its own web page? See https://github.com/kubernetes/website/pull/41793/files#diff-6837431359d119c18686bd0377481bf69921b81f4a221801d31e7fbe57aaa642 for an example of a way to do that:

Nice! I didn't know about that feature. 

Looks like a good method to pull this in and better than keeping a manual copy in repo 👍 
/approve

content/en/includes/_index.md
```diff
@@ -0,0 +1,3 @@
+---
+headless: true
```
sftim, line 2:
```suggestion
headless: true
_build:
  list: never
  render: false
```

sftim (changes requested):
Almost.
https://deploy-preview-421--kubernetes-contributor.netlify.app/includes/ exists, and I don't think it should.

A similar story for https://deploy-preview-421--kubernetes-contributor.netlify.app/includes/cncf-code-of-conduct/

sftim:
https://deploy-preview-421--kubernetes-contributor.netlify.app/includes/cncf-code-of-conduct/ still renders. Not the end of the world IMO.

/lgtm
/hold

This is an improvement and I'm OK to live with the snag that remains.

Issue #425, opened 2023-08-30
In the "Site content" section in README.md, there's a repetition of the phrase "generated from".


aadarsh-nagrath:
In the "Site content" section, there's a repetition of the phrase "generated from" in the sentence "The content for the Contributor Site is sourced from multiple locations. Content managed within this repository is generated from generated from Markdown found within the content directory." 
Consider rephrasing for clarity.
![image](https://github.com/kubernetes/contributor-site/assets/92307537/b7d7e929-3657-4d7c-a09f-fbcacdc5a15b)

aadarsh-nagrath:
On it

PR #426, opened 2023-08-30
Repetition phrase fixed

aadarsh-nagrath:
Great catch! I've addressed the duplicate wording issue in the "Site content" section. 
Issue #425


mrbobbytables:
You are congradulating yourself with "great catch" :P

FYI - For trivial fixes like this, you don't need to open an issue. You can just open a PR and fix directly.

/lgtm
/approve

PR #442, opened 2023-10-03
Adding canonical post for 2023 steering election

bridgetkromhout:
Adding correct canonical location for the blog post that went out yesterday, per @sftim as explained in https://github.com/kubernetes/website/pull/43294#issuecomment-1744076802.

Will follow up with a PR to link the kubernetes.io blog post with the right canonical url frontmatter.


sftim:
I'm wary to merge this because of the likely UX: a second thing pinging into people's feeds, 2 days after the first one showed up.

/lgtm
It's a blog article! Renders OK.

However, I'm not sure we should approve it and publish it, this time. Next time we may be able to get things more slick.

mrbobbytables:
I think it'll be okay, this site has a much smaller audience and this is where it technically should be.

PR #445, opened 2023-10-16
Finishing links for Contributor Summit

qedrakmar:
Finishing links for Contributor Summit

/assign @jberkus 
/assign @pnbrown

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -39,5 +39,6 @@
 /summit           /events/2023/kcsna  302
 /summit/reg       https://cvent.me/qRPey3 302
 /summit/cfp       https://forms.gle/htQSHpot9rp1csDz8 302
-# /summit/unconf    LINK UNCONFERENCE GITHUB ISSUE HERE 302
-# /summit/notes     LINK GOOGLE DRIVE NOTES FOLDER HERE 302
+/summit/unconf    https://github.com/kubernetes/community/issues/7531 302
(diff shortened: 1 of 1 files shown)


qedrakmar:
@jberkus I think we're good to merge.

static/_redirects
```diff
@@ -39,5 +39,6 @@
 /summit           /events/2023/kcsna  302
 /summit/reg       https://cvent.me/qRPey3 302
 /summit/cfp       https://forms.gle/htQSHpot9rp1csDz8 302
-# /summit/unconf    LINK UNCONFERENCE GITHUB ISSUE HERE 302
-# /summit/notes     LINK GOOGLE DRIVE NOTES FOLDER HERE 302
+/summit/unconf    https://github.com/kubernetes/community/issues/7531 302
+/summit/notes     https://drive.google.com/drive/folders/1plMF2WG5AKcdzQwQ5PQbmFX2xhkreUis 302
```
sftim, line 43: This is a private Google doc - is that right?

qedrakmar: It's a folder and should be shared with the org.  Can you not access it?  If so, we can look at the permissions.

jberkus: I can access it

sftim: OK, so it's meant to require a login. So long as that's intended.

mrbobbytables (commented):
/approve

Issue #448, opened 2023-10-20
Add an article on organising a Kubernetes Contributor Summit


fsmunoz:
This is an important moment for contributors and would provide some deserved exposure to the work people do to organise it.

PR #449, opened 2023-10-20
Add spotlight article on Kubernetes Contributor Summit

fsmunoz:
Add spotlight article on the Kubernetes Contributor Summit, using KCSEU2023 as the main example.

Closes: #448


jberkus:
Seems fine, unless we want to add feedback from other KCS leads?

fsmunoz:
@jberkus that would be good, yes. I did approach it as a 1:1 interview (similar to other spotlights), but it does have a broader scope.

On top of that, I'm also adding at least one photo from the event, from the ones in Flickr - I'm assuming there is no legal limitation with that usage.

fsmunoz:
@jberkus just to clarify, since after re-reading my comment I think it's a bit ambiguous: I welcome more reviewers for the content, but I wasn't aiming at changing the scope of the interview itself. We can certainly make another article with more people, even a round-table kind of thing, but I wouldn't change the structure of the current one.

fsmunoz:
/title Add spotlight article on Kubernetes Contributor Summit

fsmunoz:
/retitle Add spotlight article on Kubernetes Contributor Summit

fsmunoz:
I've added a group photo from the Flickr set.

fsmunoz:
Added https://github.com/kubernetes/website/pull/43775 for `website`.

jberkus:
Approved, that way we don't have to change the date.

sftim:
I think it's easier to line up both articles and then merge them once they're right. Maybe one day we'll have tooling to help with that.

PR #463, opened 2023-12-14
Adjust Redirects for KCSEU 2024 & Add CFP

mfahlandt:
Changing Redirects for the KCSEU 2024 Summit 

part of #464 

Resolves adding Cfp & notes task


kumarankit999 (approved):
/lgtm

PR #472, opened 2024-02-08
fix: CFP deadline extended for Kubernetes Contributor Summit Paris

fykaa:


natalisucks:
thanks for catching this fix, @fykaa! do you mind replacing the `<del>` element with `<s>`? we want to indicate that our CFP date is no longer accurate, not that information should be deleted

PR #486, opened 2024-03-19
Don't default Hugo version in Dockerfile

sftim:
For a container build, expect that this is happening wrapped in `make` (but still allow a local container image build where you specify the version of Hugo by hand)

/cc @JamesLaverack


JamesLaverack (approved):
/lgtm
/approve

PR #492, opened 2024-03-26
Make `hack/gen-content.sh` discover GNU `grep` and `sed` on macOS

embik:
Running `hack/gen-content.sh` falls apart when running on macOS if `grep` doesn't point to GNU grep due to the `-P` flag heavily being used. This is normally not the case (you _can_ install it via homebrew and override the `grep` name, but that's not the default), so I've adjusted the script to discover $GREP and $SED (just in case sed also acts up and I didn't notice) and set them to the GNU versions (which are prefixed with a g on installation). The discovery heuristics are simple but seem to be effective.

I'm not sure if this is something that is interesting to this repository (since you can also run this in a container), just wanted to submit it in case it would be useful to the community. Feel free to tell me that this will not be merged.

### DIFF
--- a/hack/gen-content.sh
+++ b/hack/gen-content.sh
@@ -47,6 +47,29 @@ if [ "${DEBUG}" != false ]; then
   VERBOSE='-v'
 fi
 
+# discover GNU grep on the system (macOS by default has BSD grep,
+# which does not support perl regex via the -P flag)
+GREP="grep"
+if $GREP --version | $GREP "BSD" >/dev/null; then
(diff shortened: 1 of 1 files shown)


stmcginnis:
I've had to do something similar to this in the past. I don't have a macOS system to test on right now, but this looks like it covers the detection and use throughout.

At any rate, testing with the changes on a Linux system doesn't show anything broken by the change, so:
/lgtm

mrbobbytables (commented):
Thanks! I appreciate the addition :)
/approve

PR #531, opened 2024-09-21
Upgrade website to Docsy 0.5.1

### LINKED ISSUE #503: Update X logo to 𝕏 on the website's footer
In the footer, we still have the old bird icon for Twitter. I suggest we update that to 𝕏.

image

### LINKED ISSUE #557: Migrate to using Docsy as an npm module
We would like to move from using Docsy as a git submodule to an NPM dependency. This is the proposed approach for the docs website (kubernetes/website#48812) and we would like to do it here too, ideally after #531 is merged.

SayakMukhopadhyay:
~~This PR updates Docsy to the latest version available and makes it compatible to use with the latest Hugo.~~

This PR was initially made to be upgraded to Docsy 0.10.0 but has since been modified to upgrade only upto 0.5.1. Further upgrade will be done in another PR. I have crossed out the items that was planned in this PR but won'
t happen in this PR either due to being fixed by another PR or being siloed to another PR.

In details, the following are the changes:

- Updated the docsy submodule to use ~~[v0.10.0 tag](https://github.com/google/docsy/releases/tag/v0.10.0)~~ [v0.5.1 tag](https://github.com/google/docsy/releases/tag/v0.5.1)
- ~~Updated dependencies in the `package.json` to the latest `autoprefixer` and `postcss-cli` and added `postcss` which is needed by `postcss-cli` now.~~
- ~~Update the Hugo version in netlify to 0.134.3 and~~ added `NODE_VERSION = 20.16.0` env var to force netlify to use the current LTS of Node.js as older versions won't be able to compile native code in some of the dependencies. **(If this value is configured on the infra side, we should look into either updating the value there and remove it from here, or remove it from infra and keep it here.)**
- Update the Makefile to add a target to download the dependencies of docsy. ~~(WIP: right now, only added as a dependency of the `preview-build` target to get the CI in the PR to work)~~
- Partials removed:
  - `head.html`: it was the same as the theme except without the `_internal/google_news.html` template. This template has since been removed.
  - ~~`navbar.htm`l: it was the same as the theme except removing the minification of the navbar logo due to Hugo not having `tdewolff/minify >= 2.7.3`. The latest Hugo has `v2.20.37` so that removal is not needed.~~
  - `sidebar-html`: it was the same as the theme except wrapping an `or` conditional method with `cond` method to work with newer versions of Hugo which Docsy was not supporting. The latest Docsy version does support the latest version of Hugo and this wrapping is no longer needed.
- Layouts removed:
  - `docs` & `layout`: this layout was an exact copy from the theme with the only difference being the `page-meta-lastmod.html` partial was removed in the override. But the partial itself doesn't activate unless both `.GitInfo` and `Site.Params.github_repo` are truthy, which is not, hence the override is not needed.
  - `events`: this layout was unnecessary as all the content files in the `events` folder are using `type: docs` in its front matter and so the `events` layout will never be used. And anyway, `docs` layout and `events` layout were exactly the same.
- Layouts updated:
  - `calendar`: updated to the current `docs` layout provided by the theme while keeping the customization of including some scripts.
  - `community`: updated to the current `docs` layout provided by the theme as it was the same before. **(The only reason this layout is needed is because all generated files in `content/en/community` doesn't contain the `type` or `layout` in its front matter and thus will default to `type: page` if a `community` layout is not created.)**
  - `resources`: updated to the current `docs` layout provided by the theme as it was the same before. **(The only reason this layout is needed is because of a generated file `content/en/resource/release/index.md` which doesn't contain the `type` or `layout` in its front matter and thus will default to `type: page` if a `resources` layout is not created.)**
- ~~`_variables.scss` has been updated from the theme. This change is introduced by [docsy updating to bootstrap 5 in 0.7.0](https://github.com/google/docsy/releases/tag/v0.7.0). This change introduces multiple changes to the default layout, including margins, padding, colors, font sizes among others.~~ `_variables.scss` has been replace by `_variables_project.scss` which is the [recommended way to customize the variables of Docsy](https://www.docsy.dev/docs/adding-content/lookandfeel/#project-style-files). This file only lists the properties that need to be customized and not everything in Docsy's `_variables.scss`
- ~~`_styles_project.scss` has been added to include the CSS to make the copyright statement visible. This is the file that all customisations should go to.~~
- ~~In the cover page, the color parameter is removed else it is setting the wrong text color.~~
- ~~The Hugo config has been changed from TOML to YAML~~ (This has been done in #556). ~~Some properties which are redundant have also been commented out. Also, since FontAwesome was upgraded to v6 in Docsy, the icon names have been updated. Also fixes #503 ~~
- Migrated from using Docsy as a Git submodule to using it as an NPM module. Fixes #557. This migration also updates the `Makefile` to not use any `git submodule` commands and update the `Dockerfile` to use `npm ci` to install the dependencies instead of installing them globally.


SayakMukhopadhyay:
The biggest "change" in this upgrade is the fact that Bootstrap was updated to v5 by Docsy. This has led to changes to stylistic elements, mostly padding, margins, text sizes and some colours. I will try to document some of them here. (Left is the current site and right is the PR site)
1. The page title is no longer capitalised by default (see https://github.com/google/docsy/pull/797) but the title and menu items are now aligned.
image
2. In the welcome page, margins are calculated based on screensize instead of being harcoded in bootstrap. Should these margins be changed to reflect the current state? Do note, the current site is using defaults. Also the link colours are a bit different.
image
3. The blog list also has very minor margin differences. Also note the slight difference in the link colours. Another difference that is present in all list pages is that the left hand side navigation looks a little different with an underline below the page title instead of the items being indented and the search box being squarish.
image
4. Disabled buttons in pagination also has a different colour
image
5. Feedback buttons look a bit different
image

These are all that I have found out. I am not too sure if there is a need to make the upgrade look pixel-accurate to the current site.

tokt:
Thank you so much!! This will take awhile to review, sorry. 
Personally: +1 on moving from toml to yaml and also on deletion of _variables.scss.

SayakMukhopadhyay:
Alright, I will bring in those changes from the other branch to make it easier to review. The config file will contain commented properties which are not needed but I have kept them for now to make it easier to review the YAML->TOML change. I can remove them before merging this PR once the format change itself is reviewed.

chris-short:
This is amazing! Thank you so much. As mentioned, we'll need some time to review all this but, I greatly appreciate the time, effort, and energy.

SayakMukhopadhyay:
I have tried to make the commits as atomic as possible so if any changes are required, we can drop those commits.

assets/scss/_variables_project.scss
```diff
@@ -0,0 +1,11 @@
+$secondary: #303030;
```
sftim, line 1: Should we set `$primary` too?

Makefile
```diff
@@ -33,6 +33,10 @@ container-targets: container-image container-gen-content container-render contai
 help: ## Show this help text.
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
 
+dependencies:
+	npm ci
+	cd themes/docsy/ && npm i
```
sftim, line 38: Does this need adapting for `make container-server`?

hugo.yaml
```diff
@@ -0,0 +1,119 @@
...
+  links:
+#   Updated icons to use latest syntax from FA.
+    user:
+      - name: Kubernetes Dev Mailing List
+        url: 'https://groups.google.com/a/kubernetes.io/group/dev'
+        icon: fa-solid fa-envelope
+        desc: Discussion and help from your fellow users
+      - name: Twitter
+        url: 'https://twitter.com/K8sContributors'
+        icon: fa-brands fa-x-twitter # Updated the bird icon to the X icon
```
sftim, line 109: Omit this comment, but once you have, comment on this line in your own PR to call out the change.

netlify.toml
```diff
@@ -3,7 +3,8 @@ publish = "public"
 command = "make production-build"
 
 [build.environment]
-HUGO_VERSION = "0.124.0"
+HUGO_VERSION = "0.134.3"
```
sftim, line 6: Is this needed? If we can do the minimal upgrade in _this_ PR, I would; save the Hugo upgrade for a different PR.

Or: split out just that change into a smaller PR that we can merge first.

sftim:
> community: updated to the current docs layout provided by the theme as it was the same before. 
> (The only reason this layout is needed is because all generated files in content/en/community doesn't contain the type or layout in its front matter and thus will default to type: page if a community layout is not created.)

Could we add that front matter via the generator? Maybe in a separate PR (which I'd prefer to merge ahead of this one)

assets/scss/_variables_project.scss
SayakMukhopadhyay: The value of `$primary` has been the same as docsy and so I didn't change it

site
https://github.com/kubernetes/contributor-site/blob/cea0531aa755544a936e69b22098f3231510f940/assets/scss/_variables.scss#L20

docsy
https://github.com/google/docsy/blob/25f73360dc4b72c7606531b37edace0417b40858/assets/scss/_variables.scss#L20

If the primary colour needs to be changed, we can set the `$primary` and maybe others too.

sftim: We should use `#326CE5` - see https://github.com/kubernetes/kubernetes/blob/master/logo/colors.md

If we don't use that already, oops, let's get a small PR merged so that we do, and then rebase / adapt this one.
I checked and we seem to use `#3176D9`. Maybe that's what early Docsy defaults to?

Makefile
SayakMukhopadhyay: Yeah, the `Makefile` needs to be updated. The `dependencies` target itself is ok but needs to be added as dependencies on `docker-server`, `container-server` and `production-build`. I have not added them as I wanted to know the sentiment regarding overhauling the `Makefile`. There is a lot of repetition (like git submodule update --init --recursive --depth 1) that can be moved to its own target and used as a dependency. Let me know if its ok to do that and I will make the changes to this PR.

sftim: Where possible, split out refactorings like that into their own (separate, smaller) PR.

assets/scss/_variables_project.scss
SayakMukhopadhyay: Sorry, I can't figure out where we use `#3176D9`. I checked and it cannot be `$primary-light` as it lightens `$primary` by 75% which is practically white when `$primary` is `#3063de`.

Makefile
SayakMukhopadhyay: Which is why I haven't made any changes to the Makefile. The `dependencies` target though is necessary for the latest Docsy and it was needed to get the `preview-build` target to work and will be needed for the other targets. Essentially, that command creates 2 folders inside `themes`, `FortAwesome` and `twbs`.

netlify.toml
SayakMukhopadhyay: The project seems to work with 0.124. So, I will revert back this change in this PR and create a new PR with the change.

SayakMukhopadhyay:
> > community: updated to the current docs layout provided by the theme as it was the same before.
> > (The only reason this layout is needed is because all generated files in content/en/community doesn't contain the type or layout in its front matter and thus will default to type: page if a community layout is not created.)
> 
> Could we add that front matter via the generator? Maybe in a separate PR (which I'd prefer to merge ahead of this one)

I plan to work on it but I was thinking of doing it in a separate PR as the current workaround works for now.

netlify.toml
SayakMukhopadhyay: Lets update hugo after this PR cause I don't know what could break if we update Hugo first. The latest version of Docsy requires a minimum version of 0.110.0 for Hugo. So, we should be able to merge this PR without changing the Hugo version.

hugo.yaml
SayakMukhopadhyay: What about the other properties that I have commented in `hugo.yaml`. I have kept those properties commented on to get some feedback, should I remove those properties altogether or should I uncomment them.

Makefile
sftim: Does this PR break local container previewing? If so, let's update the README and call out the issue in the PR description.

assets/scss/_variables_project.scss
sftim: Well, I would expect `$primary` to be `#326ce5`; can we do the small fix first and rebase this big one?

hugo.toml
sftim, line 1: I would make the move to YAML be its own separate PR.

package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.0.1",
   "description": "Kubernetes Contributor Site",
   "main": "none.js",
-  "dependencies": {},
   "devDependencies": {
-    "autoprefixer": "^9.7.6",
-    "postcss-cli": "^7.1.0"
+    "autoprefixer": "^10.4.20",
+    "postcss": "^8.4.47",
+    "postcss-cli": "^11.0.0"
```
sftim, line 9: Aside: do we need any of these outside of dev (eg for a production Netlify build)?

assets/scss/_variables_project.scss
SayakMukhopadhyay: Sure, I will create a new PR setting the new primary.

SayakMukhopadhyay: Apologies for the delay but I have created the PR for the colour at #555

Makefile
SayakMukhopadhyay: By local container previewing, do you mean using `make container-server`? It doesn't work in `master` itself. It seems there is a permission issue on the `/tmp` folder. The following error is thrown:
> ...
rsync: [receiver] chown "/tmp/src/themes/docsy/userguide/static/images/.version-banner.png.nPONje" failed: Operation not permitted (1)
rsync: [receiver] chown "/tmp/src/themes/docsy/userguide/static/images/.yes.png.iMPnje" failed: Operation not permitted (1)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.3.0]
make: *** [Makefile:80: container-server] Error 23

This issue came up in WSL and I tried a fresh setup in codespaces and the issue is the same. It seems like the issue dooesn't happen if I remove this line https://github.com/kubernetes/contributor-site/blob/f42204c6a4d640c766900b4c98c931a9d259dfb9/Makefile#L83

But running the docker commands manually works as expected and so local container previewing still works.

package.json
SayakMukhopadhyay: I don't think so. They have been `devDependencies` before the PR and stuff worked so I don't think its needed. In fact, `autoprefixer` and `postcss-cli` is installed globally in the container (which is probably why it works and is probably not how it should be).

assets/scss/_variables_project.scss
SayakMukhopadhyay: This is resolved, rebased and pushed.

hugo.toml
SayakMukhopadhyay: PR #556 created for this. Once its merged, I will rebase this PR and clean up the config a bit.

SayakMukhopadhyay: #556 has been raised for this.

SayakMukhopadhyay: #556 was merged and I have rebased this PR

sftim:
Also see https://github.com/kubernetes/website/pull/48726

sftim:
I pulled this locally and found that `make container-server` doesn't succeed. Per https://github.com/kubernetes/contributor-site/pull/531/files#r1827530755, I don't recommend merging this as-is.

Here's the error I saw:
```
Error: command error: failed to load modules: module "github.com/FortAwesome/Font-Awesome" not found in "/tmp/src/themes/github.com/FortAwesome/Font-Awesome"; either add it as a Hugo Module or store it in "/tmp/src/themes".: module does not exist
make: *** [Makefile:84: container-server] Error 1
```
My guess is that we need to ensure that the Docsy dependencies are built into the container image (this is what we're planning in https://github.com/kubernetes/website/pull/48812)

Would you like more advice on how to fix things here @SayakMukhopadhyay? You might find it easiest to bump to 0.5 first and then send in a smaller PR with the bump from 0.5 to 0.10

---

PS. Once this is good to merge, would you be willing to squash this down to 1 or 2 commits @SayakMukhopadhyay ?

Makefile
SayakMukhopadhyay: @sftim From your [comment](https://github.com/kubernetes/contributor-site/pull/531?new_mergebox=true#issuecomment-2558507697), I dunno how this convo got resolved but yeah, `container-server` target in the makefile hasn't been updated. In fact, as mentioned in the convo, all the targets needs to be updated to handle the npm dependency. See [this message](https://github.com/kubernetes/contributor-site/pull/531?new_mergebox=true#discussion_r1779454502) to know my primary concerns.

There are multiple ways this can be approached.

1. I can make the changes in the `Makefile` including adding the dependencies and can keep the refactoring for later.
2. I can add the dependencies and do the refactoring.
3. I can do the refactoring in a separate PR to be merged first and fix the dependencies here so that you can get it working.

For now, for you to test things, I will push a commit adding the dependant target to `container-server`.

SayakMukhopadhyay:
> I pulled this locally and found that make container-server doesn't succeed. Per https://github.com/kubernetes/contributor-site/pull/531/files#r1827530755, I don't recommend merging this as-is.

See my comment at https://github.com/kubernetes/contributor-site/pull/531?new_mergebox=true#discussion_r1894993238

> My guess is that we need to ensure that the Docsy dependencies are built into the container image (this is what we're planning in https://github.com/kubernetes/website/pull/48812)

The thing is a lot is happening inside the `Makefile` and `hack/gen-content.sh`. The `Dockefile` doesn't even copy over the source and instead the container runs by mounting the source into the container. This is the reason why `container-server` needs to run `npm ci`. I personally prefer copying over the source and building inside the container instead of mounting it as that leads to more reproducible builds.

> Would you like more advice on how to fix things here @SayakMukhopadhyay? You might find it easiest to bump to 0.5 first and then send in a smaller PR with the bump from 0.5 to 0.10

I was thinking of something similar. I can repurpose this PR to go up to 0.5 and create a new PR for 0.10 (or 0.11) or I can close this PR and create fresh PRs for both. I prefer keeping this PR as a lot of important convo is here.

> PS. Once this is good to merge, would you be willing to squash this down to 1 or 2 commits @SayakMukhopadhyay ?
Of course.

Makefile
SayakMukhopadhyay: Also, the rsync permission issue is still there in my system. I have pushed a commit that should get `container-server` working. If it doesn't try removing lines 87 and 88 (--cap-drop=ALL and --cap-drop=AUDIT_WRITE). Because I can't get things to work without removing them.

I don't know the real issue behind this as I assume it used to work.

sftim:
> I personally prefer copying over the source and building inside the container instead of mounting it as that leads to more reproducible builds.

It sounds like you recommend copying the entire website source code into the container image before previewing(?)

I'm not sure if you've tried contributing that way, but it'd be really unenjoyable. The idea of running a containerized preview is that you can open a local browser, make local edits and see changes reflected almost immediately.

If you instead had to quit Hugo, build a new image, and start a new server, this isn't going to feel worthwhile.

For a preview, reproducibility isn't particularly important. The canonical build environment is Netlify; we could use `netlify dev` for that, but it adds a barrier to entry. Lots of people have a local container runtime; fewer people have `netlify`.

sftim:
> I can repurpose this PR to go up to 0.5 and create a new PR for 0.10 (or 0.11)

Let's repurpose this one.

> I can close this PR and create fresh PRs for both. I prefer keeping this PR as a lot of important convo is here.

I agree, let's keep it.

SayakMukhopadhyay:
> I'm not sure if you've tried contributing that way, but it'd be really unenjoyable. The idea of running a containerized preview is that you can open a local browser, make local edits and see changes reflected almost immediately.
> 
> If you instead had to quit Hugo, build a new image, and start a new server, this isn't going to feel worthwhile.
> 
> For a preview, reproducibility isn't particularly important. The canonical build environment is Netlify; we could use `netlify dev` for that, but it adds a barrier to entry. Lots of people have a local container runtime; fewer people have `netlify`.

Ugh, you are right. I generally use other ways like dev envs to quickly integrate but build in container. I don't mind the current setup at all.

Makefile
sftim: We only support container previewing on Linux and macOS host OSs; if you're using another environment, the support may not have been added yet.

sftim:
If people want to run `hugo` locally, they can. Or they can preview in a container - as a project, we like containers. CI builds and production deploys are in Netlify and we've no plans to change that.

Makefile
SayakMukhopadhyay: Hmm...this permission issue occurs on GitHub Codespaces too. But this PR _as of now_ has a new commit that updates the `container-server` target. Can you check if that works for your local host?

I will split this PR into 2 tomorrow.

SayakMukhopadhyay:
Side note: the value of `$gray-800` used to be `#888` but was updated to `#797676` to improve accessibility. This means that even though `_variables.scss` in the project would be different from the `_variables.scss`, it's only becuase it wasn't updated. See google/docsy#540

SayakMukhopadhyay:
@sftim This PR has been modified to upgrade only upto 0.5.1. This PR also has an update to the `Makefile` to ensure that the docsy dependencies are installed on make commands. Once this PR is merged, I will work on #557 which would make things a bit simpler as far as the `Makefile` is concerned.

I can squash this PR down to a commit if you want. Let me know.

Makefile
```diff
@@ -62,19 +66,19 @@ docker-gen-content:
...
 	git submodule update --init --recursive --depth 1
 	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hugo --verbose --ignoreCache --minify
 
-docker-server:
+docker-server: dependencies
 	@echo -e "**** The use of docker-server is deprecated. Use container-server instead. ****" 1>&2
 	$(MAKE) container-server
 
-container-server: ## Run Hugo locally within a container, available at http://localhost:1313/
+container-server: dependencies ## Run Hugo locally within a container, available at http://localhost:1313/
```
sftim, line 81: This is actually:
```diff
-container-server: dependencies ## Run Hugo locally within a container, available at http://localhost:1313/
+container-server: dependencies ## Run npm locally on your host computer; then run Hugo locally within a container, available at http://localhost:1313/
```

I'd prefer to see the approach from k/website where NPM doesn't run locally; one of the benefits of a container is isolating less trusted code, whereas this approach executes it on the host.

assets/scss/_variables_project.scss
```diff
@@ -0,0 +1,2 @@
+$primary: #326ce5;
```
sftim, line 1: I'm surprised to see this change as part of the PR; consider rebasing against _main_?

SayakMukhopadhyay: See https://www.docsy.dev/docs/adding-content/lookandfeel/#project-style-files

sftim: Ah, that explains it. Using `_variables_project.scss` is correct.

Makefile
SayakMukhopadhyay: I have done it this way as other non-container targets would need to do this on the host anyway. Moreover, even `container-server` executes `git submodule update --init --recursive --depth 1` in the host which too could be done in the container. I am following that style in this PR.

This will need to be reworked anyway when we migrate from submodules to npm modules and I suggest that this change be done then.

sftim: We can assume that people trust `git`, but we should minimize the other code they need to trust.

I'm really reluctant to go the other way and widen what contributors must trust.

SayakMukhopadhyay: @sftim I have made the changes. This PR now also closes #557 and all `npm` commands run inside the container and not on the host.

SayakMukhopadhyay: My only question now is for the `render` and `server` commands. Since these 2 commands run on the host anyway, do you think it's ok to add `npm ci` to these commands? Or should the fact that `npm ci` needs to be run manually be added to the docs and we leave it up to the folks who clone this repo to run `npm` commands?

sftim: Yes, those are fine to run on the host.

sftim:
The container image doesn't build for me:
```
STEP 5/8: RUN npm ci
npm ERR! code EUSAGE
npm ERR!
npm ERR! The `npm ci` command can only install with an existing package-lock.json or
npm ERR! npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or
npm ERR! later to generate a package-lock.json file, then try again.
npm ERR!
npm ERR! Clean install a project
npm ERR!
npm ERR! Usage:
npm ERR! npm ci
npm ERR!
npm ERR! Options:
npm ERR! [--install-strategy <hoisted|nested|shallow|linked>] [--legacy-bundling]
npm ERR! [--global-style] [--omit <dev|optional|peer> [--omit <dev|optional|peer> ...]]
npm ERR! [--include <prod|dev|optional|peer> [--include <prod|dev|optional|peer> ...]]
npm ERR! [--strict-peer-deps] [--foreground-scripts] [--ignore-scripts] [--no-audit]
npm ERR! [--no-bin-links] [--no-fund] [--dry-run]
npm ERR! [-w|--workspace <workspace-name> [-w|--workspace <workspace-name> ...]]
npm ERR! [-ws|--workspaces] [--include-workspace-root] [--install-links]
npm ERR!
npm ERR! aliases: clean-install, ic, install-clean, isntall-clean
npm ERR!
npm ERR! Run "npm help ci" for more info

npm ERR! A complete log of this run can be found in: /root/.npm/_logs/2024-12-24T12_50_27_990Z-debug-0.log
Error: building at STEP "RUN npm ci": while running runtime: exit status 1
make: *** [Makefile:55: container-image] Error 1
```

Try cloning your repo (you can pull from a filesystem clone to save bandwidth) and check that a container image build runs OK from that clean clone @SayakMukhopadhyay 

Once it does, feel free to squash to 1 or 2 commits.

SayakMukhopadhyay:
@sftim would it be ok, if I `COPY` over the `package.json` and `package-lock.json` over to the `Dockerfile`. I could do `npm i autoprefixer ...` again but I think its redundant to list down the required packages in the `Dockerfile` when they are already there in the `package.json`.

sftim:
> would it be ok, if I `COPY` over the `package.json` and `package-lock.json` over to the `Dockerfile`.

Sure, that's a really common technique for this situation.

SayakMukhopadhyay:
@sftim It's fixed. Do you think there is any value in checking in a CI step that the container builds and runs?

sftim:
> Do you think there is any value in checking in a CI step that the container builds and runs?

One for a follow-up PR, and maybe best if we involve SIG Testing folks.

sftim:
Please squash this down to fewer commits; I think that's tidiest.

sftim (commented):
This is clear progress.

Make sure that, once you've built an image, `make container-server` runs without an error.

Warnings are OK, but it's good to fix those too.

SayakMukhopadhyay:
Alright, the couple of warnings have been addressed. I will squash this PR to 3 commits, 1 containing all the upgrades, 1 containing the submodule -> npm change and the last containing the misc changes.

Dockerfile
```diff
@@ -18,10 +18,9 @@ RUN apk add --no-cache \
 
 WORKDIR /src
 
-# Required for PostCSS
-RUN npm install -G \
-    autoprefixer \
-    postcss-cli
+COPY package*.json ./
+
+RUN npm ci
```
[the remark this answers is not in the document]
SayakMukhopadhyay: Welp....that's right. Maybe I should move the `npm ci` into `hack/gen-content.sh`. 

Seems like k8s/website too has this issue.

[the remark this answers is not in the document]
SayakMukhopadhyay: In fact, in the website, `npm ci` is in the `Dockerfile` (see https://github.com/kubernetes/website/blob/30350f69b1b1f258a4534be4e16f966a641b4f35/Dockerfile#L44). The way this works is that in the `Makefile`, each directory is mounted explicitly (see https://github.com/kubernetes/website/blob/main/Makefile#L16) and the whole directory isn't mounted as a whole.

So, we have 2 ways to work this.

1. We can use the website way of doing things.
2. I can remove the `npm ci` from the `Dockerfile` and put it in the `Makefile` as
```
container-server: ## Run Hugo locally within a container, available at http://localhost:1313/
	# no build lock to allow for read-only mounts
	$(CONTAINER_RUN) -p 1313:1313 \
		--mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 \
		--read-only \
		--cap-drop=ALL \
		--cap-drop=AUDIT_WRITE \
		$(CONTAINER_IMAGE) \
	bash -c 'cd /src && hack/gen-content.sh --in-container && \
		 cd /tmp/src && \
		 npm ci && \
		hugo server \
		--environment preview \
		--logLevel info \
		--noBuildLock \
		--bind 0.0.0.0 \
		--buildDrafts \
		--buildFuture \
		--disableFastRender \
		--ignoreCache \
		--destination /tmp/hugo \
		--cleanDestinationDir'
```
(note the `npm ci && \`).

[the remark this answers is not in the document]
sftim: > We can use the website way of doing things.

Let's do that. Otherwise, `npm` is still running in the host context, and the trust base for our [static site generation website preview] code becomes much larger.

sftim:
This doesn't yet look right @SayakMukhopadhyay.

Here's the error I get when I try a local preview:
```console
make container-server
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
# no build lock to allow for read-only mounts
docker run --user : --rm -it -p 1313:1313 \
	--read-only --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/.git,target=/src/.git,readonly --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/assets,target=/src/assets,readonly --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/content,target=/src/content,readonly --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/external-sources,target=/src/external-sources,readonly --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/hack,target=/src/hack,readonly --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/layouts,target=/src/layouts,readonly --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/static,target=/src/static,readonly --mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 --mount type=bind,source=/home/tim/Documents/Repositories/kubernetes/contributor-site/hugo.yaml,target=/src/hugo.yaml,readonly \
	--cap-drop=ALL \
	--cap-drop=AUDIT_WRITE \
	k8s-contrib-site-hugo \
bash -c 'cd /src && hack/gen-content.sh --in-container && \
	 cd /tmp/src && \
	hugo server \
	--environment preview \
	--logLevel info \
	--noBuildLock \
	--bind 0.0.0.0 \
	--buildDrafts \
	--buildFuture \
	--disableFastRender \
	--ignoreCache \
	--destination /tmp/hugo \
	--cleanDestinationDir'
Copying source repository
rsync: [generator] chown "/tmp/src/." failed: Operation not permitted (1)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.3.0]
make: *** [Makefile:92: container-server] Error 23
```

SayakMukhopadhyay:
@sftim as discussed in Slack, the issue is due to the `--user :` argument which is forcing the container to run as root and thus causing some permission issues.  The PR #421 adds the `--user  :` argument probably due to some issue with podman. The website doesn't use the --user argument and it runs fine.

I have also tried running podman myself without the `--user :` argument and it works with the latest podman so I believe whatever the issue was has been solved by podman.

I have added a commit which removes `--user :` so could you please try once again.

P.S. we should add a CI job to build and test run the container.

SayakMukhopadhyay:
PR rebased

Dockerfile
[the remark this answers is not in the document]
sftim: Looks resolved

SayakMukhopadhyay:
@sftim squashed to 2 commits, 1 with the docsy upgrade and the other with the npm module upgrade.

sftim (commented):
Local time is 16:59 Friday

_this is fine_

/lgtm
/approve

chris-short:
Great work everyone!

tokt:
+1 and kudos to @SayakMukhopadhyay  and @sftim, this was epic!!

sftim:
Successor PR: https://github.com/kubernetes/contributor-site/pull/563

PR #539, opened 2024-10-09
Update KCS faq and KCS registration page that KubeCon registration is required

reylejano:
This PR adds 2 additions about KubeCon registration requirements:
- an FAQ entry that KubeCon registration is required
- a line on the Registration page that KubeCon registration is required


stmcginnis:
/lgtm

Thanks for doing this @reylejano! I think this should help cut down on at least some of the questions.

PR #555, opened 2024-12-15
fix: update the primary colour to align with the k8s branding

SayakMukhopadhyay:
Updated the primary colour of the site based on the k8s branding as discussed below.    
    
    Well, I would expect `$primary` to be `#326ce5`; can we do the small fix first and rebase this big one?

_Originally posted by @sftim in https://github.com/kubernetes/contributor-site/pull/531#discussion_r1827532103_


sftim (commented):
Thanks

/lgtm
/approve

PR #556, opened 2024-12-16
feat: move from a toml hugo config to a yaml one

SayakMukhopadhyay:
Moving to YAML config for Hugo instead of TOML as Docsy itself has moved. This PR will also be a blocker for #531 as this change will be removed from that PR and rebased.
Reference [comment](https://github.com/kubernetes/contributor-site/pull/531?new_mergebox=true#discussion_r1827532753) by @sftim 
> I would make the move to YAML be its own separate PR.

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -1,159 +1,164 @@
-baseURL = "https://www.kubernetes.dev/"
-title = "Kubernetes Contributors"
-theme = ["docsy"]
-enableRobotsTXT = true
-enableGitInfo = false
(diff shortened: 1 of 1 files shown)


jberkus:
Wow, thanks!

Netlify preview looks good, what could this break that wouldn't be immediately obvious, if anything?

SayakMukhopadhyay:
I don't think anything would break as I have converted the TOML to YAML verbatim including the comments.

PR #561, opened 2025-01-11
refactor: remove pygments and other properties that are already default

### LINKED ISSUE #560: Remove pygments from hugo config file
pygments configuration is no longer needed for Docsy (see https://github.com/google/docsy/discussions/2150). This is true for the latest version of Docsy but I don't know how true it is for older versions.

So, this might only be worked on after at least PR #531 is merged.

SayakMukhopadhyay:
Updated `hugo.yaml` removing some properties which are deprecated and or the same as the defaults.
- Remove pygments in config which closes #560 which includes `pygmentsCodeFences`, `pygmentsUseClasses`, `pygmentsUseClassic`, `pygmentsStyle`.
- Remove `enableGitInfo` as `false` is the default, see https://gohugo.io/getting-started/configuration/#enablegitinfo
- Remove `defaultContentLanguage` as `en` is the default, see https://gohugo.io/getting-started/configuration/#defaultcontentlanguage
- Remove `defaultContentLanguageInSubdir` as `false` is the default, see https://gohugo.io/getting-started/configuration/#defaultcontentlanguageinsubdir
- Remove `imaging.quality` as `75` is the default, see https://gohugo.io/content-management/image-processing/#quality
- Remove `imagine.anchor` as `smart` is the default, see https://gohugo.io/content-management/image-processing/#anchor
- Remove `privacy_policy` as an empty string does the same thing as not having the property
- Remove `version_menu`, `archived_version`, `version` as the site is not versioned.
- Remove `algolia_docsearch` and `offlineSearch` as only 1 method of searching can be enabled and Algolia search and offline search is disabled by default. See https://github.com/google/docsy/blob/e6d94771f3376c1f5f96c3be3f01759b529cdc8d/layouts/partials/head.html#L60 and https://www.docsy.dev/docs/adding-content/search/
- Remove `breadcrumb_disable` as `false` is the default. Breadcrumbs are enabled by default see https://www.docsy.dev/docs/adding-content/navigation/#breadcrumb-navigation
- Remove `sidebar_search_disable` as `false` is the default. Sidebar search is enabled by default see https://www.docsy.dev/docs/adding-content/search/#disabling-the-sidebar-search-box
- Remove `navbar_logo` as `true` is the default. Navbar logo is enabled by default see https://www.docsy.dev/docs/adding-content/iconsimages/#add-your-logo
- Use YAML references in `languages.en.title` as its already defined in `title`.
- Add `markup.highlight.style` as `emacs` to make the style the same as k8s.io. Although nothing is using highlight shortcode in the site right now.

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -1,43 +1,31 @@
 baseURL: 'https://www.kubernetes.dev/'
-title: Kubernetes Contributors
+title: &title Kubernetes Contributors # Create a reference to use later
 theme:
   - docsy
 themesDir: node_modules
 enableRobotsTXT: true
-enableGitInfo: false
 
 # Language settings
 contentDir: content/en
-defaultContentLanguage: en
-defaultContentLanguageInSubdir: false
-enableMissingTranslationPlaceholders: true
 
+enableMissingTranslationPlaceholders: true
 disableKinds:
   - taxonomy
-
-# Highlighting config
-pygmentsCodeFences: true
-pygmentsUseClasses: false
-pygmentsUseClassic: false
-pygmentsStyle: tango
 
 # Configure how URLs look like per section.
 permalinks:
   blog: '/:section/:year/:month/:day/:slug/'
 
-
 # Image processing configuration.
 imaging:
   resampleFilter: CatmullRom
-  quality: 75
-  anchor: smart
 
 services:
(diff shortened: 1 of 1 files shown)


sftim:
Thanks

/lgtm
/approve

PR #568, opened 2025-01-29
Footer Social Icons

chris-short:
1. Removed X/Twitter
2. Added LinkedIn and Mastodon to footer
3. Updated the FontAwesome dependency to add Bluesky but could not get the Bluesky icon to display despite the code being in the HTML output

Signed-off-by: Chris "Not So" Short <chrisshort@duck.com>

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -148,10 +148,18 @@ params:
         url: 'https://groups.google.com/a/kubernetes.io/group/dev'
         icon: fa fa-envelope
         desc: Discussion and help from your fellow users
-      - name: Twitter
-        url: 'https://twitter.com/K8sContributors'
-        icon: fab fa-twitter
-        desc: Follow us on Twitter to get the latest news!
+      - name: LinkedIn
+        url: 'https://www.linkedin.com/company/kubernetes/'
+        icon: fab fa-linkedin
+        desc: Follow us on LinkedIn to get the latest news!
+      - name: Mastodon
+        url: 'https://hachyderm.io/@K8sContributors'
+        icon: fab fa-mastodon
+        desc: Follow us on Mastodon to get the latest news!
+      - name: Bluesky
+        url: 'https://bsky.app/profile/kubernetes.dev'
+        icon: fab fa-bluesky
+        desc: Follow us on Bluesky to get the latest news!
 # Developer relevant links. These will show up on right side of footer and in the community page if you have one.
     developer:
       - name: GitHub
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -156,10 +156,6 @@ params:
         url: 'https://hachyderm.io/@K8sContributors'
         icon: fab fa-mastodon
         desc: Follow us on Mastodon to get the latest news!
-      - name: Bluesky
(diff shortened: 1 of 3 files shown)


SayakMukhopadhyay:
I had tried doing the same, upgrading FontAwesome doesn't work cause Docsy FA dependency is shadowing the explicit dependency.

chris-short:
Bluesky is going to take some extra work but, @sftim this is ready for review along with https://github.com/kubernetes/website/pull/49594

hugo.yaml
```diff
@@ -148,10 +148,18 @@ params:
...
+        icon: fab fa-linkedin
+        desc: Follow us on LinkedIn to get the latest news!
+      - name: Mastodon
+        url: 'https://hachyderm.io/@K8sContributors'
+        icon: fab fa-mastodon
+        desc: Follow us on Mastodon to get the latest news!
+      - name: Bluesky
+        url: 'https://bsky.app/profile/kubernetes.dev'
+        icon: fab fa-bluesky
+        desc: Follow us on Bluesky to get the latest news!
```
sftim, line 162: Should we merge this before the Bluesky icon works? I'm unsure.

package.json
```diff
@@ -5,6 +5,7 @@
   "devDependencies": {
     "autoprefixer": "^9.7.6",
     "docsy": "github:google/docsy#semver:0.6.0",
+    "@fortawesome/fontawesome-free": "^6.7.2",
```
SayakMukhopadhyay, line 8: Do you think that this can be removed as updating fontawesome doesn't help?

hugo.yaml
chris-short: One less thing to change later. I'm indifferent.

package.json
chris-short: Might as well, removed

SayakMukhopadhyay: Apologies for the nitpick, but the lock file change should be reverted too.

SayakMukhopadhyay:
Apologies for the nitpick, but could you revert the lock file too?

chris-short:
Done! @SayakMukhopadhyay @sftim

hugo.yaml
sftim: If you drop it, I can definitely merge the change @chris-short. I feel like adding a broken link to Bluesky is technical debt (and potentially a source of issues, etc).

chris-short: Done.

hugo.yaml
```diff
@@ -148,10 +148,14 @@ params:
         url: 'https://groups.google.com/a/kubernetes.io/group/dev'
         icon: fa fa-envelope
         desc: Discussion and help from your fellow users
-      - name: Twitter
-        url: 'https://twitter.com/K8sContributors'
-        icon: fab fa-twitter
-        desc: Follow us on Twitter to get the latest news!
```
sftim, line 154: Why this removal? I get that Twitter is losing market share and the geopolitics but is that strong enough grounds to make the change?

chris-short: After the comms meeting it appears my definition of deprioritize is too heavy-handed.

PR #584, opened 2025-06-16
Fix hyperlink syntax in Changes to Kubernetes Slack

lmktfy:
Use Markdown syntax for the hyperlink


lmktfy:
I'm happy that https://github.com/kubernetes/website/pull/51258 can merge even before this fix has merged to the original article.

PR #587, opened 2025-07-14
Use current GitHub username

lmktfy:
Update my identity: @sftim → @lmktfy


tokt:
/approve

*crossed fingers it'll work

lmktfy:
(also needs LGTM)

PR #602, opened 2025-11-22
chore: Upgrade Hugo and adjust footer links

TineoC:
This PR upgrades the Hugo version to 0.152.2, which brings performance improvements, new features, and bug fixes. It also cleans up temporary files introduced during the modernization effort

- Moves the LinkedIn link from the developer section to the user section and updates its icon.
- Adjusts the footer link ordering.


TineoC:
@chris-short, hey Chris, I tried upgrading to the 0.152 version. We could follow up in the next call your thoughts :)

PR #625, opened 2026-01-20
fix(header): add background on scroll to prevent content overlap

### LINKED ISSUE #624: Sticky header overlaps content during scroll — needs background for readability
## Description

On the page `https://kubernetes.dev/welcome/`, when scrolling down, the sticky header becomes visible but has no background color. This causes the white text of the header to overlap with the underlying content, making it hard to read and visually confusing.

The issue is not present on initial load — only when scrolling.

## Expected Behavior

When the header becomes sticky during scroll, it should have a solid or semi-transparent background (e.g., blue like the footer) to ensure contrast and readability.

## Screenshots

Image
Image
Image
Image

## Environment

- Browser: Chrome / Firefox / Safari
- OS: macOS / Windows / Linux

## Suggested Fix

Add a background color (e.g., `#1a75ff`) to the header when it becomes sticky. Optionally add `backdrop-filter: blur(10px)` for a modern look.

Example CSS:
```css
.header.scrolled {
  background-color: #1a75ff;
  backdrop-filter: blur(10px);
}

### PR DESCRIPTION
**Pull request by codeurluce**
Fixes #624

When scrolling down, the sticky header currently has no background, causing underlying text to appear through it — reducing readability.

This PR:
- Adds a semi-transparent Kubernetes blue background (`rgba(26, 117, 255, 0.95)`) to `.td-navbar` when scrolled
- Uses a lightweight scroll listener to toggle the `.scrolled` class
- Loads the script via Docsy's `body-end.html` hook

No build or Go knowledge required — pure frontend fix.

/kind bug

### CODE REVIEW

File: assets/scss/_styles_project.scss [hunk 1]
```diff
@@ -1,3 +1,11 @@
 .navbar-brand__name {
   text-transform: uppercase;
 }
+
+.td-navbar {
+  transition: background-color 0.3s ease;
+}
+
+.td-navbar.scrolled {
+  background-color: rgba(26, 117, 255, 0.95);
```
lmktfy: You might be able to reference `$primary` here (I haven't checked it this works), at 50% alpha.

lmktfy (commented):
Thanks for the PR.

It's very tidy; rarely do I see a first contribution that is so well aligned to how we work. 

Until I properly test this I won't add LGTM, but I am happy to have this change merged.

PR #647, opened 2026-02-05
feat(build): migrate content ingestion to hugo modules

### LINKED ISSUE #93: Update method in which external content is ingested.

TineoC:
### Summary
Migrates the external content ingestion workflow from the custom `hack/gen-content.sh` script to **Hugo Modules**. This change modernizes the build process and leverages Hugo's native support for mounting content from external Git repositories.

Content freshness is handled entirely within Netlify's build pipeline: production builds run `hugo mod get -u` before building, and a daily scheduled deploy ensures content stays up to date even when no PRs merge. This eliminates the need for external automation (Prow / GitHub Actions).

### Fixes
Fixes #93

/sig contribex
/cc @SayakMukhopadhyay @reylejano


Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apk add --no-cache \
     build-base \
     curl \
     git \
+    go \
```
lmktfy: I would add a version pin or otherwise try to ensure it is close to the version that Netlify uses.

go.sum
lmktfy: Under what circumstances will these values change?

lmktfy (commented):
Thanks for the PR.

I like the ambition, a lot, but I have concerns about the detail.

Makefile
```diff
@@ -28,17 +28,23 @@ CONTAINER_IMAGE			:= $(IMAGE_REPO):$(GIT_TAG)
...
 	--mount type=bind,source=$(CURDIR)/assets,target=/src/assets,readonly \
 	--mount type=bind,source=$(CURDIR)/content,target=/src/content,readonly \
-	--mount type=bind,source=$(CURDIR)/external-sources,target=/src/external-sources,readonly \
-	--mount type=bind,source=$(CURDIR)/hack,target=/src/hack,readonly \
 	--mount type=bind,source=$(CURDIR)/layouts,target=/src/layouts,readonly \
 	--mount type=bind,source=$(CURDIR)/static,target=/src/static,readonly \
 	--mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 \
 	--mount type=bind,source=$(CURDIR)/hugo.yaml,target=/src/hugo.yaml,readonly
+# Writable mount for container-render output (Hugo writes to /out -> host public/)
+CONTAINER_RENDER_MOUNT	:= --mount type=bind,source=$(CURDIR)/public,target=/out
```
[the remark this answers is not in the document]
TineoC: Thanks, good catch

.github/workflows/update-content.yml
```diff
@@ -0,0 +1,54 @@
...
+          commit-message: "chore: update external content modules"
+          title: "Update external content modules"
+          body: |
+            This PR updates the Hugo modules to pull the latest content from upstream repositories.
+            
+            This is an automated update. Prow will merge it after CI passes.
+            
+            /sig contributor-experience
+            /kind feature
+            /lgtm
```
lmktfy: Will this work? You usually can't LGTM your own PR.

lmktfy (commented):
To be honest, I wouldn't use GitHub Actions for this. Is it feasible to either:

* not commit when a source repository changes, and calculate things during a site build
* extend Prow to automate the new branch submission and checking, rather than use GitHub Actions

?

go.sum
lmktfy: I would have automated builds, and notify maintainers if no utomated builds have succeeded in the last 36 hours.

lmktfy: > we also have some security/stability risks because we also run the risk of breaking the site every time there's new content in every other repo and page.

If such a change breaks the build, the existing site still gets served.

TineoC:
> To be honest, I wouldn't use GitHub Actions for this. Is it feasible to either:
> 
> * not commit when a source repository changes, and calculate things during a site build
> * extend Prow to automate the new branch submission and checking, rather than use GitHub Actions
> 
> ?

This looks good to you?

https://github.com/kubernetes/test-infra/pull/36933

go.sum
TineoC: What if we just run https://github.com/kubernetes/contributor-site/pull/647/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R75 `make module-download` in dev and make that only the `prod` environment is the one making sure it is running with the latest go modules and in dev and feature branches, keeping the ones tracked by git in the go.mod file?

@lmktfy @jberkus

TineoC:
/close

Closing this pr as I'd rather follow-up splitting this changes

Issue #590, opened 2025-09-08
For events page, a link to new Maintainer Summits


pacoxu:
refer to https://github.com/cncf/contribute-site/issues/33.

Until now, we don't have a perfect link for following Maintainer Summits. (remind me if I don't know)

kaslin:
@pacoxu @TineoC Maintainer Summits are owned by the CNCF, but we do still aim to keep the community informed about them. The best page for learning about a maintainer summit would be the one on the associated KubeCon's page, ie. https://events.linuxfoundation.org/kubecon-cloudnativecon-europe/features-add-ons/maintainer-summit/
Granted that this will change each KubeCon, I don't think there's a consistent link for up-to-date Maintainer Summit info. And going forward, there are no guarantees that folks in the Kubernetes community would have all the info to maintain such a page, at least not without specifically designating someone to try.
With Contributor Summits, we could add the responsibility of updating the website to the role handbook for comms lead. We're still debating if we will have similar roles for each Maintainer Summit.
FYI @mfahlandt .

PR #666, opened 2026-02-24
fix: enable future date pages in contributor-site

TineoC:
Future builds aren't available, so it is ignoring future date pages.

Currently, is showing the meet & greet as it doesn't have a date on its page.

image

https://www.kubernetes.dev/events/2026/kcseu/

Relates to https://github.com/kubernetes/contributor-site/issues/590

CC @kaslin @stmcginnis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ gen-content: ## Generates content from external sources.
 	hack/gen-content.sh
 
 render: dependencies ## Build the site using Hugo on the host.
-	hugo --logLevel info --ignoreCache --minify
+	hugo --logLevel info --ignoreCache --minify --buildFuture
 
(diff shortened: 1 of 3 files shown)


stmcginnis:
Awesome, nice work @TineoC!

/lgtm
/approve

Makefile
```diff
@@ -175,7 +175,8 @@ production-build: ## Builds the production site (this command used only by Netli
 		--environment production \
 		--logLevel info \
 		--ignoreCache \
-		--minify
+		--minify \
+		--buildFuture
```
lmktfy: This is a bug. We should not publish future dated content (it will make blog articles go live too early).

TineoC: Thanks for pointing that out. I think I should create a pr to move our events "date" to use another property like "eventDate" and turn future builds off.

PR #671, opened 2026-03-01
fix: revert global buildFuture to fix future blog post publication

### LINKED ISSUE #669: Netlify build incorrectly builds future content
PR #666 incorrectly sets **production** builds to include future content. This will break when we next have an article to publish.

Prior to #666, the site was designed to use scheduled builds and to take blog articles live by waiting until they were in the past. The first production build after publication date would take it live.

We should fix that.
@TineoC I _may_ have misunderstood, but I think we should partially revert #666 and find another way to represent the date for upcoming events, so that they can be included even as we avoid publishing future dated content in general.

/kind bug
/priority important-soon

TineoC:
This PR reverts the changes introduced in #666 to address #669.

### Description
PR #666 correctly enabled building of future events, but it did so by setting `buildFuture: true` globally in `hugo.yaml` and adding the flag to production build targets in the `Makefile`. This caused future-dated blog posts to be published immediately in production, breaking the scheduled publication mechanism.

Fixes #669

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ gen-content: ## Generates content from external sources.
 	hack/gen-content.sh
 
 render: dependencies ## Build the site using Hugo on the host.
-	hugo --logLevel info --ignoreCache --minify --buildFuture
+	hugo --logLevel info --ignoreCache --minify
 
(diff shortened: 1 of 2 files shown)


SayakMukhopadhyay:
This PR reverts the change by #666
/lgtm

PR #672, opened 2026-03-01
feat(content): normalize event dates to ensure build visibility

TineoC:
This PR ensures that future event pages remain visible and correctly sorted after the global `buildFuture` configuration is reverted in #671. 

### Changes
- Introduces a custom `eventDate` field for events to ensure visibility in non-future builds.
- Adds a custom `section-index.html` partial to maintain correct chronological sorting by the new `eventDate` property.

Depends on: #671
cc @lmktfy

### DIFF
--- a/content/en/events/2020/_index.md
+++ b/content/en/events/2020/_index.md
@@ -1,5 +1,6 @@
 ---
 title: 2020
+eventDate: 2020-01-01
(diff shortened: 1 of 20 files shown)


chris-short:
#671 has been merged.

/unhold

PR #693, opened 2026-04-03
Security: Hardening GitHub Actions workflows and configuration

TineoC:
Added an explicit `permissions: contents: none` block to the `build` job. Since this job only triggers a webhook via `curl`, it does not require access to the repository's source code or other resources.

/sig contributor-experience


lmktfy (commented):
/lgtm

PR #700, opened 2026-04-11
Draft: Implement Community Resilience Dashboard for SIG ContribEx

TineoC:
This PR implements the Community Resilience Dashboard (Lottery Factor) as requested in #681.

### Features
- **Go Ingestion Engine**: Automatically discovers SIG ContribEx repositories, parses OWNERS for maintainers, and detects tech stacks using GitHub API and file heuristics.
- **Interactive Treemap**: A custom Hugo shortcode using Apache ECharts to visualize knowledge distribution.
- **Project Detail Modals**: Deep metadata including Maintainers (Approvers/Reviewers), Tech Stack badges (Hugo, Docsy, Prow, etc.), and entry points for contributors.
- **Automation**: GitHub Action to refresh data weekly.

Relates to #681


hack/gen-lottery-factor.go
```diff
@@ -0,0 +1,418 @@
...
+	outLangs, _ := exec.Command("gh", "api", fmt.Sprintf("repos/%s/languages", repo)).Output()
+	var langs map[string]int
+	json.Unmarshal(outLangs, &langs)
+
+	for l := range langs {
+		name := l
+		if l == "Dockerfile" { name = "Docker" }
+		if l == "Shell" { name = "Bash" }
+		if l == "Markdown" { name = "Documentation" }
+		if l == "YAML" { name = "Infrastructure" }
```
lmktfy: (nit) not sure how strong this link is

data/lottery_factor.json
lmktfy, line 1: This could go into Git as `data/lottery_factor.json` and then be rendered out at build time.

Personally, I would do that. It makes the data available to Hugo, which we might find useful.

assets/js/lottery-treemap.js
```diff
@@ -0,0 +1,199 @@
...
+      }
+
+      const formattedData = data.subprojects.map(sub => {
+        const subData = {
+          name: sub.name,
+          children: sub.repos.map(repoName => {
+            const stats = data.repo_data.find(r => r.repo === repoName);
+            if (!stats) return null;
+
+            // Map Lottery Factor to color
```
lmktfy: Can we make this accessible to people who have trouble distinguishing colors?

assets/js/lottery-treemap.js
```diff
@@ -0,0 +1,199 @@
+(function () {
+  const chartDom = document.getElementById('lotteryTreemap');
+  if (!chartDom) return;
+  const myChart = echarts.init(chartDom);
+
+  fetch('/data/lottery_factor.json')
```
lmktfy: The URL makes sense to me.

layouts/shortcodes/lottery-factor.html
```diff
@@ -0,0 +1,29 @@
+<div class="lottery-factor-container my-4">
+  <div id="lotteryTreemap" style="width: 100%; height: 600px; border-radius: 8px; border: 1px solid #dee2e6; box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075);"></div>
```
lmktfy: nit: I like to put the CSS separated from HTML content.

layouts/shortcodes/lottery-factor.html
```diff
@@ -0,0 +1,29 @@
...
+      <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
+      </div>
+    </div>
+  </div>
+</div>
+
+<script src="https://cdn.jsdelivr.net/npm/echarts@5.4.3/dist/echarts.min.js"></script>
+
+{{ $lotteryJs := resources.Get "js/lottery-treemap.js" | minify | fingerprint }}
```
lmktfy: nit: it's debatable, but the way I like to load this kind of script is from the page end layout hook (`body-end.html`) using a Has shortcode conditional.

layouts/shortcodes/lottery-factor.html
```diff
@@ -0,0 +1,29 @@
...
+      <div class="modal-header">
+        <h5 class="modal-title" id="projectInfoModalLabel">Project Details</h5>
+        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        <div id="projectInfoBody">
+           <!-- Content will be injected by JS -->
+        </div>
+      </div>
+      <div class="modal-footer">
```
lmktfy: Inject the close button too? (makes it a bit nicer if JS isn't available).

hack/gen-lottery-factor.go
```diff
@@ -0,0 +1,418 @@
...
+		LotteryFactor:      lf,
+		TotalPoints:        total,
+		Contributors:       contributors,
+		LastUpdated:        time.Now().Format(time.RFC3339),
+		TechStack:          techStack,
+		Owners:             owners,
+		OnboardingURL:      onboardingUrl,
+		OwnersURL:          ownersUrl,
+		IssuesURL:          fmt.Sprintf("https://github.com/%s/issues", repo),
+		GoodFirstIssuesURL: fmt.Sprintf("https://github.com/%s/issues?q=is%%3Aissue+is%%3Aopen+label%3A%%22good+first+issue%%22", repo),
```
lmktfy: nit: I would sort these oldest-first.

lmktfy (commented):
I think there are snags with doing the analysis, because the things that have a lottery factor are SIGs and working groups (and SIG subprojects). But the analysis is centered on repos.

How do we plan to tackle that mismatch?

hack/gen-lottery-factor.go
TineoC: Not at all, I was discussing with Kaslin if we could ask SIG leads to provide the tech stacks in the sigs.yaml file. Because pulling this with GitHub metadata would just provide information of the programming languages of the projects.

We have to decide if we would like to handle this static or dynamically

content/en/community/resilience.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Understanding the Visualization
+- **Size of boxes:** Represents the total activity (points) in that subproject/repository.
+- **Color:** Indicates the Lottery Factor.
+    - <span style="color: #dc3545; font-weight: bold;">Red (LF <= 2):</span> High Risk - Single Point of Failure.
+    - <span style="color: #ffc107; font-weight: bold;">Yellow (LF 3-4):</span> Moderate Risk.
+    - <span style="color: #28a745; font-weight: bold;">Green (LF >= 5):</span> Healthy distribution.
+
+### Take Action
+If you see a project with a low Lottery Factor, it's a great opportunity to get involved!
```
tokt: just a nitpick: Would it be possible to rephrase this as "with a high risk" or something like that? 
It's not immediately obvious that "low Lottery Factor" = High Risk - SPOF.

Issue #701, opened 2026-04-11
Upgrade Nodejs version for production environment


TineoC:
The production environment is running on Nodejs 12

<img width="1148" height="639" alt="Image" src="https://github.com/user-attachments/assets/3c8728d4-5000-4f12-9627-07c0f9f8af3a" />

Let's upgrade this to the LTS currently.

Madhurchandran:
Hi @TineoC,
Before I open a PR, wanted to confirm the target version. The current Active LTS is Node 24 (v24.16.0), but PR #704 previously targeted Node 20, which is now in Maintenance LTS.
Given the jump from Node 12, I'm considering three options:

Node 22 — Active LTS (Maintenance), very stable, broad ecosystem support.
Node 24 — current Active LTS, recommended for production.
Node 20 — as in PR #704, but now entering end of its Maintenance phase.

thank you

TineoC:
After we fix the master branch I would like to get this in place https://github.com/kubernetes/contributor-site/issues/736 so for further upgrades we're preventing this issues.

TineoC:
@Madhurchandran also, we would like to mirror the same node and Hugo versions as kubernetes/website so we might first discuss this with SIG docs if they are in the roadmap for upgrading to a newer node version first

Madhurchandran:
Hi @TineoC, I completely agree that aligning with kubernetes/website is the best approach to ensure compatibility and stability for Hugo builds and related tooling. I'm interested in this discussion and would be happy to stay involved as the roadmap and versioning plans are discussed with SIG Docs.

Thank you

PR #704, opened 2026-04-24
chore: align contributor Node version metadata

abhay1999:
## Summary
- add `.nvmrc` and `package.json` engine metadata for Node.js `20.16.0`
- document the expected Node.js version in the README
- align local contributor setup with the Node.js version already used by Netlify production builds

## Context
`netlify.toml` already pins `NODE_VERSION = "20.16.0"`, so this PR focuses on making that version explicit for local development and contributor workflows as follow-up cleanup for issue #701.

## Testing
- not run (metadata and documentation change only)

Refs #701


TineoC:
image

Is weird that none of our env vars in hugo.yaml are overriding this value.

@lmktfy do you have experience with this?

[From their docs:
](https://docs.netlify.com/build/configure-builds/manage-dependencies/?_gl=1*gc7z2h*_gcl_au*MTQyMjk2Nzk2NS4xNzcwMjczMTEyLjkxMTg1ODY1NS4xNzc3NDAyMDQ3LjE3Nzc0MDI2NTg.)
> Note that a NODE_VERSION environment variable, .node-version file, or .nvmrc file will override this UI setting.

abhay1999:
Thanks for flagging this, @TineoC! You're right that `netlify.toml` already sets `NODE_VERSION = "20.16.0"` under `[build.environment]`, but per the Netlify docs you quoted, `.nvmrc` / `.node-version` files take even higher precedence. This PR's `.nvmrc` acts as a stronger enforcement mechanism on top of what's already in `netlify.toml` — it doesn't change the target version (still `20.16.0`), it just makes it harder for the UI or any other setting to drift. If the env var in `netlify.toml` isn't being picked up reliably, this `.nvmrc` should resolve that.

jberkus:
Deploy looks fine, any reason not to merge the cleanup?

TineoC:
@abhay1999 is possible to re-use the value in https://github.com/kubernetes/contributor-site/blob/master/netlify.toml#L6 so we don't have to keep track of our versions in more than one file?

PR #707, opened 2026-04-29
Render KEP README content on k8s.dev

### LINKED ISSUE #698: Render KEP Readme instead of a custom template in k8s.dev
Do all KEPs follow the same structure as in https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/19-Graduate-CronJob-to-Stable (for example)?

I think our current approach: https://www.kubernetes.dev/resources/keps/19/

It was a good start, but we're not displaying any actual information of what this KEP is doing in the website, we're basically redirecting people to github so it could actually know what it is about.

I'm I wrong?

I would love to hear some thoughts on this.

CC. @kubernetes/sig-contributor-experience-pr-reviews

TineoC:
This PR implements the rendering of full KEP README content directly on k8s.dev.

Currently, KEP pages only show metadata and redirect users to GitHub. This change fetches the README content during the build process and displays it prominently on the page.

Fixes https://github.com/kubernetes/contributor-site/issues/698

Special notes: This PR was prepared with the help of Antigravity.

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -28,34 +28,234 @@ body.td-home div.lead p {
 
 /* KEP page styles */
 .kep-page {
-  .kep-stage-badge {
-    margin-bottom: 1rem;
+  padding: 1rem 0;
 
-    .badge {
-      font-size: 0.9rem;
-      padding: 0.5em 1em;
-      border-radius: 4px;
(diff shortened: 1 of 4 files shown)


jberkus:
Huge improvement here.  And it seems to handle KEPs that are malformed OK by just not displaying the README content?

/lgtm

PR #709, opened 2026-04-30
SEO and Core Web Vitals optimizations

TineoC:
Closes #685

This PR implements comprehensive SEO and performance optimizations for kubernetes.dev:
- **Core Web Vitals:** Defers render-blocking scripts, adds LCP preloads with high fetch priority, and uses swap font display.
- **Canonical URLs:** Points canonical tags directly to production.
- **Meta Descriptions:** Standardizes and injects description fields in the frontmatter of all tracked documentation and blog pages.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -19,6 +19,7 @@ This file needs to be removed once Docsy has been updated to 0.10.0.
 <meta name="robots" content="noindex, nofollow">
 {{ end -}}
 
+<link rel="canonical" href="{{ .Permalink }}">
 {{ partialCached "favicons.html" . }}
 <title>
   {{- if .IsHome -}}
@@ -36,7 +37,7 @@ This file needs to be removed once Docsy has been updated to 0.10.0.
 <script
   src="https://code.jquery.com/jquery-3.6.0.min.js"
   integrity="sha384-vtXRMe3mGCbOeY7l30aIg8H9p3GdeSe4IFlP6G8JMa7o7lXvnz3GFKzPxzJdPfGK"
-  crossorigin="anonymous"></script>
+  crossorigin="anonymous" defer></script>
 {{ if .Site.Params.offlineSearch -}}
 <script defer
   src="https://unpkg.com/lunr@2.3.9/lunr.min.js"
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -37,7 +37,7 @@ This file needs to be removed once Docsy has been updated to 0.10.0.
 <script
   src="https://code.jquery.com/jquery-3.6.0.min.js"
   integrity="sha384-vtXRMe3mGCbOeY7l30aIg8H9p3GdeSe4IFlP6G8JMa7o7lXvnz3GFKzPxzJdPfGK"
-  crossorigin="anonymous" defer></script>
+  crossorigin="anonymous"></script>
 {{ if .Site.Params.offlineSearch -}}
 <script defer
   src="https://unpkg.com/lunr@2.3.9/lunr.min.js"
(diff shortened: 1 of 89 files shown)


layouts/partials/head.html
```diff
@@ -36,7 +37,7 @@
 <script
   src="https://code.jquery.com/jquery-3.6.0.min.js"
   integrity="sha384-vtXRMe3mGCbOeY7l30aIg8H9p3GdeSe4IFlP6G8JMa7o7lXvnz3GFKzPxzJdPfGK"
-  crossorigin="anonymous"></script>
+  crossorigin="anonymous" defer></script>
```
TineoC: <COMMENT>
[HIGH] Deferring jQuery while having inline scripts (like those in `body-end.html` or `calendar/baseof.html`) that use `$` will cause `ReferenceError: $ is not defined`. Standard inline scripts execute immediately during parsing, whereas deferred scripts execute later.

To fix this, consider removing `defer` from jQuery or wrapping all dependent inline scripts in a `window.addEventListener('DOMContentLoaded', ...)` block (though even then, they must wait for the deferred script to execute).
</COMMENT>

layouts/partials/scripts.html
```diff
@@ -0,0 +1,125 @@
+{{ $needKaTeX  := or .Site.Params.katex.enable .Params.math .Params.chem -}}
```
TineoC, line 1: <COMMENT>
[MEDIUM] Overriding the entire `scripts.html` partial from Docsy adds a maintenance burden, as updates to the theme's script management will need to be manually ported here. 

While necessary to add `defer` to these specific scripts, consider if Docsy provides hooks or smaller partials that could be overridden instead to achieve the same result more surgically.
</COMMENT>

layouts/shortcodes/blocks/cover.html
```diff
@@ -0,0 +1,52 @@
...
+{{ $col_id := .Get "color" | default "dark" -}}
+{{ $image_anchor := .Get "image_anchor" | default "smart" -}}
+{{ $logo_anchor := .Get "logo_anchor" | default "smart" -}}
+{{/* Height can be one of: auto, min, med, max, full. */ -}}
+{{ $height := .Get "height" | default "max" -}}
+
+{{ with $promo_image -}}
+{{/* Use WebP for better performance and smaller size */ -}}
+{{ $promo_image_big := (.Fill (printf "1920x1080 webp %s" $image_anchor)) -}}
+{{ $promo_image_small := (.Fill (printf "960x540 webp %s" $image_anchor)) -}}
```
TineoC, line 14: <COMMENT>
[LOW] Excellent use of `fetchpriority="high"` and WebP conversion for the hero image. This is a highly effective optimization for improving the Largest Contentful Paint (LCP) metric.
</COMMENT>

TineoC (commented):
## 📋 Review Summary

The Pull Request successfully introduces essential SEO and performance optimizations, specifically targeting Core Web Vitals and URL canonicalization. While the implementation of `fetchpriority` and WebP is highly commendable, the deferral of jQuery introduces a critical risk of breaking existing site functionality that relies on inline scripts.

## 🔍 General Feedback

- **Necessity**: The changes are highly necessary for a high-traffic production site. Canonical tags prevent duplicate content issues, while LCP optimizations directly impact search ranking and user experience.
- **404 Investigation**: I have reviewed the changes and found no logic that would cause 404 errors for KEP pages. These files primarily affect the `<head>` and script loading, not the routing or content generation logic. If 404s are occurring, they likely stem from the earlier README ingestion changes or a local build cache issue.
- **Script Dependencies**: The primary concern is the dependency chain of jQuery. Moving to `defer` is the right goal, but requires a holistic update to all inline scripts.

jberkus:
I am not qualified to review this, I'm afraid.

PR #739, opened 2026-06-09
pin SHA hash for gcb-docker-gcloud image

Prajyot-Parab:
- pin SHA hash for gcb-docker-gcloud


Priyankasaggu11929:
/lgtm
/approve

Thank you!

PR #735, opened 2026-05-30
fix: malformed href attributes in homepage and event FAQ pages

immanuwell:
Fixes broken links in the homepage and 6 event FAQ pages.

Issue 1 - `content/en/_index.md`: three `href` values have stray whitespace (one has both leading and trailing spaces), one uses `http://` instead of `https://`. Browsers include the spaces in the URL so those links just 404.

Issue 2 - `content/en/events/{2022,2023,2024}/*/faq.md` (6 files): the `kubernetes-csi` list item is missing a closing `"` after the URL. Browser reads `https://github.com/kubernetes-csi rel="noopener noreferrer" target="_blank"` as the literal href value -- link is completely busted.

How to repro: open https://www.kubernetes.dev/ and hover "developer guides", "Contributor Experience meetings", "becoming a reviewer or approver", or "governance" -- spaces show up in the URL bar. For the FAQ issue open any 2022-2024 event FAQ and click the kubernetes-csi org link.

Refs #679

Issue #729, opened 2026-05-25
Failing deploys following Hugo version bump


lmktfy:
I am not sure, but it seems to me that after https://github.com/kubernetes/contributor-site/pull/728 merged, the Netlify deploy didn't succeed.

Really I've filed this issue to mark that we ought to check that things are OK for this website. I don't have enough access to Netlify to check on that myself.

/kind bug

lmktfy:
Also, we never got a deploy preview for #728 (I think Netlify might be misconfigured) and Prow is happy to merge a PR even without seeing a successful deploy preview (a second misconfiguration?)

Prajyot-Parab:
Fix - https://github.com/kubernetes/community/pull/9003

Prajyot-Parab:
Fix - https://github.com/kubernetes/contributor-site/pull/733

Prajyot-Parab:
Netlify prod builds should probably be fixed by above fixes, still need to debug and fix preview build (failing due to node_modules issue)

PR #749, opened 2026-06-13
fix: replace curly quotes with straight quotes in etcd-io link

TineoC:
The href and target attributes on the etcd-io list item in events/2024/kcseu/faq.md used curly typographic quotes (U+201C/U+201D) instead of standard straight quotes. Browsers do not recognize curly quotes as HTML attribute delimiters, so both href and target were ignored, making the link completely non-functional.

This complements the kubernetes-csi link fixes in PR #735.

Refs #679


TineoC:
Closing this PR — it contains 3 out-of-scope commits from PR #729 (charter_link handling). Will recreate with only the etcd-io fix.

PR #766, opened 2026-06-13
Upgrade Docsy from 0.8.0 to 0.9.1

Madhurchandran:
## Summary

Upgrade Docsy from 0.8.0 to 0.9.1 as the first step towards enabling dark mode support on contributor-site.

## Changes
- Bump Docsy from v0.8.0 to v0.9.1
- Regenerate and update package-lock.json to align with the upgraded dependency tree
- Add a local mermaid.html partial override to address a Mermaid rendering/build issue introduced during the upgrade
- Rebase onto the latest master and reconcile upgrade-related template changes
- Validate KEP listing and detail page rendering against recent upstream KEP layout and accessibility updates
- Exclude TBD milestone values from the KEP release filter generation logic

## Testing
- Ran make server
- Verified successful local site build
- Validated KEP listing and detail page rendering
- Spot-checked page rendering after the upgrade

## Related

Ref: #745

### AI Disclosure 
An AI assistant was used to help investigate the Resource.Err compatibility issue. All proposed changes were reviewed, tested, and validated by me before submission.

PR #783, opened 2026-06-16
Modularize SCSS into partials and centralize color tokens

Caesarsage:
Restructures `assets/scss/` to mirror kubernetes/website's organization. No visual change — light mode only; compiled CSS is value-identical.

Split`_styles_project.scss` into focused partials (`_navbar`, `_home`, `_community`, `_kep_page`, `_kep_filter`, `_kep_table`, `_base`, `_reset`) and centralize repeated colors into tokens in `_variables_project.scss` (neutral ramp + KEP stage palette). Each token resolves to its original hex.

Foundation for dark mode: with colors centralized as tokens and styles modular, future dark-mode work overrides tokens instead of editing inline hexes across one file.

#766 **Whichever merges first, the other needs a rebase** to reconcile those two files. 
The changes are complementary (both lead toward dark mode); this PR rewrites both files wholesale, so the rebase is mostly re-applying the smaller delta either way.


lmktfy:
Thanks

/lgtm
/approve

PR #788, opened 2026-06-19
Remove mt-5 from homepage welcome paragraph

TineoC:
Removes the Bootstrap `mt-5` margin-top class from the welcome paragraph on the homepage.


stmcginnis:
Can you include the reason for removing it? I can't see a difference in the preview, nor historical reasons why it was part of the header, but it does [look like a pretty common thing](https://cs.k8s.io/?q=mt-5&i=nope&literal=nope&files=&excludeFiles=&repos=).

chris-short:
There is extra white space in the dark gray block. Removing `mt-5` improves the page's flow. I'm good with this change.

/lgtm

/approve

PR #812, opened 2026-06-28
chore: consolidate Node version into netlify.toml only

### LINKED ISSUE #807: Consolidate Node versions so that they are set in only one place
Per [this comment](https://github.com/kubernetes/contributor-site/pull/704#issuecomment-4632186500), we are currently setting the Node version in two, maybe three places, which means that they get out of sync.  It would be good to change things to set them in only one place.

/help

TineoC:
Consolidate all Node version references into netlify.toml.

Fixes: #807


TineoC:
[netlify.toml](https://github.com/kubernetes/contributor-site/blob/main/netlify.toml#L6) is using 20.17 that's why the version is changing

.nvmrc
Prajyot-Parab, line 1: Do we need to track this file? Since it will always be generated, it probably shouldn't be committed. If that's the case, we should add it to the `.gitignore` file.

Issue #830, opened 2026-07-29
Netlify periodic build hook still points to deleted 'master' branch


TineoC:
## Problem

Periodic Netlify build trigger (cron, `.github/workflows/netlify-periodic-build.yml`) are failling after the [project was moved to `main`.](https://github.com/kubernetes/contributor-site/issues/686)

## Root cause

Site's only build hook (`xxxx`, id `xxxxx`) is bound to branch `master`. The repo's default branch was renamed to `main`.

## Fix (needs Netlify site admin + GitHub repo secret admin)

1. New build hook scoped to `main`:
   `netlify api createSiteBuildHook --data '{"site_id":"xxxxxx","title":"periodic-main","branch":"main"}'`
2. Point the secret at it:
   `gh secret set NETLIFY_BUILD_HOOK_KEY --repo kubernetes/contributor-site --body "<new-hook-id>"`
3. Delete the stale hook:
   `netlify api deleteSiteBuildHook --data '{"site_id":"xxxxxx","build_hook_id":"xxxxx"}'`
4. Verify: trigger the workflow manually, confirm a new deploy for the latest `main` shows in Netlify's deploy log.

TineoC:
cc. @kubernetes/owners

Priyankasaggu11929:
x-posting, should be fixed now - https://github.com/kubernetes/org/issues/6518#issuecomment-5112203839

Also, @TineoC, let's be careful and not add any credentials ( `build_hook_id`) in the PR description.

For now, I have deleted the exposed `build_hood_id` for master branch.

Priyankasaggu11929:
@TineoC, is this good to close now?

TineoC:
yes!

/close

PR #831, opened 2026-07-29
fix: fail netlify periodic-build workflow loudly on hook failure

TineoC:
## What
`.github/workflows/netlify-periodic-build.yml` used `curl -s` to hit the Netlify build hook, discarding the HTTP response. Non-2xx responses were swallowed, so the scheduled job reported green even when no build was actually triggered.

## Why
Relates to #830 the periodic build pointing to `master` (which doesnt exist anymore) was failing, but the GH actions were never failing in the logs.

### DIFF
--- a/.github/workflows/netlify-periodic-build.yml
+++ b/.github/workflows/netlify-periodic-build.yml
@@ -11,6 +11,6 @@ jobs:
       contents: none
     steps:
     - name: Trigger webhook on netlify
-      run: curl -sS --fail-with-body -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
+      run: curl -sS -f -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
       env:
         TOKEN: ${{ secrets.NETLIFY_BUILD_HOOK_KEY }}
--- a/.github/workflows/netlify-periodic-build.yml
+++ b/.github/workflows/netlify-periodic-build.yml
@@ -11,6 +11,6 @@ jobs:
       contents: none
     steps:
     - name: Trigger webhook on netlify
-      run: curl -s -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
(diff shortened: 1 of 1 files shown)


TineoC:
```
❯ curl --help all | grep fail
 -f, --fail                                        Fail fast with no output on HTTP errors
     --fail-early                                  Fail on first transfer error
     --fail-with-body                              Fail on HTTP errors but save the body
     --mail-rcpt-allowfails                        Allow RCPT TO command to fail
~ ❯                                                                                   Py base Node system
```

.github/workflows/netlify-periodic-build.yml
```diff
@@ -11,6 +11,6 @@ jobs:
       contents: none
     steps:
     - name: Trigger webhook on netlify
-      run: curl -s -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
+      run: curl -sS --fail-with-body -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
```
Priyankasaggu11929: Thank you for catching it and for the PR @TineoC!

Can we make sure that enabling this flag doesn't output too much information which is "not intended to be publicly available" in the logs?

TineoC: I removed `fail-with-body` as we don't care about the body here; I just want to give an exit code different than 0 when it finishes.

I can't test this locally, but passing -S and -f, so in case of a non-zero should be captured by GitHub Actions

no error; expected output

```
(nothing else — step ends green, exit 0)
```

error condition:

```
curl: (22) The requested URL returned error: <code >
##[error]Process completed with exit code <code>.
```

Priyankasaggu11929:
/lgtm
/approve

Thank you again!

TineoC:
Cool, I see the scheduled builds are triggered correctly

image

PR #838, opened 2026-08-05
fix: keep KEP detail card columns balanced

### LINKED ISSUE #837: style: kep detail card overflow
Image

We need a max width on the KEP Detail card so it's split on half, ideally

from: https://www.kubernetes.dev/resources/keps/6060/

TusharChauhan09:
Fixes uneven column widths in the KEP detail metadata card. Closes:  #837 

The KEP metadata banner uses a two-column CSS grid, but long content in the "Related Enhancements" section could influence the intrinsic size of the first column. As a result, the Implementation History column expanded while the Ownership column became noticeably narrower, making the card look unbalanced on desktop and medium-sized screens.

CSS only, scoped to `.kep-page` in `assets/scss/_kep_page.scss`. No markup or content changes.

Changes:
- Use `minmax(0, 1fr)` for the metadata banner grid so both columns share the available width evenly.
- Allow banner columns to shrink correctly by adding `min-width: 0`.
- Allow long Related Enhancement links to wrap naturally using `overflow-wrap: anywhere`.
- Replace the flex layout for Related Enhancement links with a block layout so wrapped links no longer affect the column width calculation.
- Keep the leading icon aligned by replacing `flex-shrink` with a right margin.

Verified locally with the pinned Hugo v0.144.2 (`make container-server`). The KEP pages build successfully and the metadata banner maintains balanced columns.

- Deploy preview :
https://deploy-preview-838--kubernetes-contributor.netlify.app/resources/keps/6060/

## Before

KEP detail card before fix — uneven columns, long URLs overflow into Ownership column

## After

KEP detail card after fix — balanced columns, URLs wrap in left column

### DIFF
--- a/assets/scss/_kep_page.scss
+++ b/assets/scss/_kep_page.scss
@@ -12,7 +12,7 @@
 
   .kep-banner-grid {
     display: grid;
-    grid-template-columns: 1fr 1fr;
+    grid-template-columns: minmax(0, 1fr) minmax(0, 1fr);
 
     @media (max-width: 991px) {
       grid-template-columns: 1fr;
@@ -22,6 +22,7 @@
   .kep-banner-col {
     padding: 2rem;
     position: relative;
+    min-width: 0;
 
     &.border-start-refined {
       border-left: 1px solid #e0e4e8;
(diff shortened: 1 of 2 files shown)


assets/scss/_kep_page.scss
```diff
@@ -228,18 +229,15 @@
         font-size: 0.85rem;
         color: $slate-700;
         text-decoration: none;
-        display: flex;
```
TineoC, line 231: is possible to fix it keeping the flex display?

TusharChauhan09: Yeah, kept display: flex on the link items and fixed the alignment with some other flex properties instead. Pushed the latest commits, let me know if it looks good now

TusharChauhan09: found one more issue - long URLs were still overflowing with display : flex. So I kept it like you asked, just wrapped the URL text in a span so it could actually shrink and wrap

TineoC:
I see https://www.kubernetes.dev/resources/keps/6060/ is already fixed.

Also noted in the preview the icons aren't centered in the [preview](https://deploy-preview-838--kubernetes-contributor.netlify.app/resources/keps/6060/) link 
image

PR #805, opened 2026-06-25
Upgrade Docsy from 0.9.1 to 0.10.0

Madhurchandran:
## Summary
This PR begins the incremental upgrade of Docsy from 0.9.1 to 0.10.0, following the staged upgrade path toward newer Docsy releases and support for features such as dark mode.

## Changes
- Upgrade Docsy from 0.9.1 to 0.10.0.
- Update npm dependencies and package-lock.json.
- Pin the Mermaid version via params.mermaid.version.
- Update the Mermaid partial to use try (resources.GetRemote) with proper error handling for CDN retrieval failures.

## Current Status
 This is a PR and is still under active development.

## Related
 Ref: https://github.com/kubernetes/contributor-site/issues/745
 
### AI Disclosure
An AI assistant was used to help identify and fix issues during regression testing. All changes were reviewed and validated before submission.

PR #855, opened 2026-08-10
fix: own the RSS template and emit a single blog feed

### LINKED ISSUE #854: RSS feeds re-announcing archived posts and non-blog pages after Docsy 0.10.0 upgrade
### What happened

Since the Docsy 0.10.0 upgrade landed on Saturday, the site's RSS feeds changed shape and started re-announcing old content. Downstream automation picked it up — the [2020 site launch announcement](https://www.kubernetes.dev/blog/2020/08/24/announcing-the-contributor-website/) was auto-posted to X and LinkedIn this week, and personal reposting automation surfaced assorted archived pages.

Nothing was republished on the site. `content/en/blog/2020/announcing-the-contributor-website.md` has not been touched since 2020 (`b63aee0`).

### Root cause

Docsy 0.10.0 removed `layouts/_default/list.rss.xml` — "Drop RSS layout in favor of Hugo's default" in the [v0.10.0 release notes](https://github.com/google/docsy/releases/tag/v0.10.0) (google/docsy#1948).

This repo owns no feed template of its own, so every `index.xml` was rendered by Docsy's. #805 merged as 0647fe9 on 2026-08-09, and Hugo 0.144.2 silently fell back to [`_internal/rss.xml`](https://github.com/gohugoio/hugo/blob/v0.144.2/tpl/tplimpl/embedded/templates/_default/rss.xml). The build does not warn — output just changes.

Hugo's internal template selects pages with no section filter:

```go-html-template
{{- if .IsHome }}{{ $pctx = .Site }}{{ end }}
{{- if or $.IsHome $.IsSection }}
{{- $pages = $pctx.RegularPages }}
```

and only caps items when `services.rss.limit` is set, which this site does not set.

### Impact

Measured locally against `origin/main` with Docsy 0.10.0 installed:

| Feed | With Docsy 0.9.1 template | Current (Hugo built-in) |
|---|---|---|
| `/index.xml` items | 50, blog only | **802**, every page on the site |
| `/index.xml` with `01 Jan 0001` pubDate | 0 | **733** |
| `/blog/index.xml` items | 50 | 67 |
| item title | `Blog: <title>` | `<title>` |

The 733 zero-dated entries are docs, events, awards and charter pages, which carry no `date` front matter.

The 17 items newly exposed in `/blog/index.xml` are the oldest posts — 2020-08-24 through 2022-08-15, previously cut off by the `first 50` cap. Guids were unchanged by the upgrade, so subscribers saw them as genuinely new entries. That is the set that got auto-posted.

### Why kubernetes/website is unaffected

kubernetes/website has never used the theme's feed template. It owns its own and restricts `outputs` so only the home page emits RSS — and both predate Docsy on that site by years:

- [`layouts/index.rss.xml`](https://github.com/kubernetes/website/blob/main/layouts/index.rss.xml) was added in [`7f3b633a`](https://github.com/kubernetes/website/commit/7f3b633a), "Convert site to Hugo" (#8316, 2018-05-05). Only 5 commits have ever touched it; the last content change was [`6ebcdd1f`](https://github.com/kubernetes/website/commit/6ebcdd1f) (2023-12-21).
- The [`[outputs]` block in `hugo.toml`](https://github.com/kubernetes/website/blob/main/hugo.toml#L84-L88) carries the comment "We (currently) only want an RSS feed for the home page". `git blame` attributes lines 84–85 and 87 to that same 2018 commit; line 88 was amended by [`7bc3a7d8`](https://github.com/kubernetes/website/commit/7bc3a7d8) (2021-03-07) to add `print`, and line 86 by [`f6b6444f`](https://github.com/kubernetes/website/commit/f6b6444f) (2025-10-29) to add `REDIRECTS`.

There was nothing for the Docsy change to take away from them.

### Fix

Adopt the same arrangement here — see #855. Own `layouts/index.rss.xml` (copied from kubernetes/website, only the `<description>` differs) and restrict `outputs` so RSS is emitted for the home page alone. `/blog/index.xml` redirects to `/index.xml` so existing subscribers keep working.

Docsy stays on 0.10.0. This is not a rollback, and no config-only fix exists:

- Docsy still declares `params.rss_sections: [blog]` at [every tag through v0.16.0](https://github.com/google/docsy/blob/v0.16.0/theme/hugo.yaml#L8), but nothing has read it since the layout was deleted.
- `services.rss.limit` caps item count but cannot filter sections.

No Docsy release after 0.10.0 restores an RSS layout — verified against the git tree of every tag through v0.16.0.

The new feed was diffed against a rebuild of the pre-incident Docsy 0.9.1 feed: same 50 guids, same order, identical `pubDate` and `<description>` on every item. Only the `Blog: ` title prefix is dropped, which matches what is live today, so nothing re-fires.

### Follow-up (separate)

`/docs/index.xml` filtered by `Type`, and all 70 pages under `content/en/events/` set `type: docs` — so event and award pages landed in the docs feed while `/events/index.xml` rendered empty. #855 removes those feeds outright, so the symptom is gone, but the `type:` mismatch in front matter is still worth correcting.

Also worth reporting upstream that dropping the RSS layout is a silent, subscriber-visible regression for theme consumers.

TineoC:
Fixes #854

### Problem

Docsy shipped `layouts/_default/list.rss.xml` through v0.9.1 and [removed it in v0.10.0](https://github.com/google/docsy/releases/tag/v0.10.0) (google/docsy#1948). This site owned no feed template, so when #805 merged (0647fe9, 2026-08-09) every feed fell back to Hugo's [`_internal/rss.xml`](https://github.com/gohugoio/hugo/blob/v0.144.2/tpl/tplimpl/embedded/templates/_default/rss.xml) — no section filter, no item cap.

### Fix

Adopt what kubernetes/website has done since [`7f3b633a`](https://github.com/kubernetes/website/commit/7f3b633a) (2018): own the template, emit RSS for the home page only. That repo was never affected because it never used the theme's.

- **`layouts/index.rss.xml`** — copied from [theirs](https://github.com/kubernetes/website/blob/main/layouts/index.rss.xml); only `<description>` differs. Filters to blog and caps at 50 in the template, not via a theme param — Docsy still declares `rss_sections` but nothing has read it since the layout was deleted, and no release through v0.16.0 restores one.
- **`hugo.yaml`** — `outputs` restricted, mirroring [their `hugo.toml`](https://github.com/kubernetes/website/blob/main/hugo.toml#L84-L88). Drops 73 dateless section feeds. `print`/`HEADERS`/`REDIRECTS` omitted; unused here.
- **`static/_redirects`** — `/blog/index.xml` → `/index.xml`, so subscribers keep working.

Docsy stays on 0.10.0.

### DIFF
--- a/layouts/index.rss.xml
+++ b/layouts/index.rss.xml
@@ -1,0 +1,56 @@
+{{/*
+  Home-page RSS feed, copied from kubernetes/website's layouts/index.rss.xml:
+  https://github.com/kubernetes/website/blob/main/layouts/index.rss.xml
+  Only the <description> text differs, since it names this site's blog.
+
+  This site owns its feed template rather than inheriting one from the theme.
+  Docsy shipped layouts/_default/list.rss.xml through v0.9.1 and removed it in
+  v0.10.0 (google/docsy#1948); with no local template, Hugo fell back to its
+  built-in RSS, which ranges over every regular page with no section filter and
+  no item cap. kubernetes/website has never been exposed to that, having owned
+  this file since it moved to Hugo in 2018:
+  https://github.com/kubernetes/website/commit/7f3b633a
+
+  Paired with the `outputs` block in hugo.yaml, which emits RSS for the home
+  page only, mirroring:
+  https://github.com/kubernetes/website/blob/main/hugo.toml#L84-L88
+  Section feeds are deliberately not generated.
+*/}}
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
+  <channel>
+    <title>{{ site.Title }} Blog</title>
+    <link>{{ .Permalink }}</link>
+    <description>The Kubernetes contributor blog is used by the project to communicate community reports, contributor news, and any updates that might be relevant to people contributing to Kubernetes.</description>
+    <generator>Hugo -- gohugo.io</generator>{{ with site.LanguageCode }}
+    <language>{{.}}</language>{{end}}{{ with site.Author.email }}
+    <managingEditor>{{.}}{{ with site.Author.name }} ({{.}}){{end}}</managingEditor>{{end}}{{ with site.Author.email }}
+    <webMaster>{{.}}{{ with site.Author.name }} ({{.}}){{end}}</webMaster>{{end}}{{ with site.Copyright }}
+    <copyright>{{.}}</copyright>{{end}}{{ if not .Date.IsZero }}
+    <lastBuildDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>{{ end }}
+    <image>
+      <url>https://raw.githubusercontent.com/kubernetes/kubernetes/master/logo/logo.png</url>
+      <title>The Kubernetes project logo</title>
+      <link>{{ .Permalink }}</link>
+    </image>
+    {{ with .OutputFormats.Get "RSS" }}
+    {{ printf "<atom:link href=%q rel=\"self\" type=%q />" .Permalink .MediaType | safeHTML }}
+    {{ end }}
+    {{ range first 50 (where site.RegularPages "Type" "in" (slice "blog")) }}
+    <item>
+      <title>{{ .Title }}</title>
+      <link>{{ .Permalink }}</link>
+      <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
+      {{ with site.Author.email }}<author>{{.}}{{ with site.Author.name }} ({{.}}){{end}}</author>{{end}}
+      <guid>{{ .Permalink }}</guid>
+      <description>
+        {{ $img := (.Resources.ByType "image").GetMatch "*featured*" }}
+        {{ with $img }}
+        {{ $img := .Resize "640x" }}
+        {{ printf "<![CDATA[<img src=\"%s\" width=\"%d\" height=\"%d\"/>]]>" $img.Permalink $img.Width $img.Height | safeHTML }}
+        {{ end }}
+        {{ .Content | html }}
+      </description>
+    </item>
+    {{ end }}
+  </channel>
+</rss>
--- a/layouts/index.rss.xml
+++ b/layouts/index.rss.xml
@@ -1,7 +1,11 @@
 {{/*
-  Home-page RSS feed, copied from kubernetes/website's layouts/index.rss.xml:
+  Home-page RSS feed, inspired from kubernetes/website's layouts/index.rss.xml:
   https://github.com/kubernetes/website/blob/main/layouts/index.rss.xml
-  Only the <description> text differs, since it names this site's blog.
+
+  The item count is read from services.rss.limit in hugo.yaml rather than
+  hardcoded. That key is 0 when unset and Hugo treats `first 0` as unlimited,
+  so the fallback below keeps a missing key from restoring the whole-site feed
+  this template exists to prevent.
 
   This site owns its feed template rather than inheriting one from the theme.
   Docsy shipped layouts/_default/list.rss.xml through v0.9.1 and removed it in
@@ -35,7 +39,9 @@
     {{ with .OutputFormats.Get "RSS" }}
     {{ printf "<atom:link href=%q rel=\"self\" type=%q />" .Permalink .MediaType | safeHTML }}
     {{ end }}
-    {{ range first 50 (where site.RegularPages "Type" "in" (slice "blog")) }}
+    {{- $limit := site.Config.Services.RSS.Limit }}
+    {{- if lt $limit 1 }}{{ $limit = 50 }}{{ end }}
+    {{ range first $limit (where site.RegularPages "Type" "in" (slice "blog")) }}
     <item>
       <title>{{ .Title }}</title>
       <link>{{ .Permalink }}</link>
--- a/layouts/index.rss.xml
+++ b/layouts/index.rss.xml
@@ -54,7 +54,7 @@
         {{ $img := .Resize "640x" }}
         {{ printf "<![CDATA[<img src=\"%s\" width=\"%d\" height=\"%d\"/>]]>" $img.Permalink $img.Width $img.Height | safeHTML }}
         {{ end }}
-        {{ .Content | html }}
+        {{ .Content | transform.XMLEscape | safeHTML }}
       </description>
     </item>
     {{ end }}
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -16,6 +16,20 @@ disableKinds:
 permalinks:
   blog: '/:section/:year/:month/:day/:slug/'
 
+# Be explicit about the output formats. We only want an RSS feed for the home
+# page, rendered by layouts/index.rss.xml. Leaving RSS on `section` would emit a
+# feed for every section, none of which carry dates, and leaving it unset lets
+# Hugo's built-in template range over every page on the site.
+# Mirrors https://github.com/kubernetes/website/blob/main/hugo.toml#L84-L88
+outputs:
+  home:
+    - HTML
+    - RSS
+  page:
+    - HTML
+  section:
+    - HTML
+
 # Image processing configuration.
 imaging:
   resampleFilter: CatmullRom
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -38,6 +38,9 @@ services:
   googleAnalytics:
     # Fake ID in support of [params.ui.feedback]. The real GA ID is set in the Netlify config.
     id: UA-00000000-0
+  rss:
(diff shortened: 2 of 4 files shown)


stmcginnis:
/lgtm

Just to make sure everyone is ready:

/hold

TineoC:
### Reproduce

```bash
# 1. live baseline
curl -sS https://www.kubernetes.dev/index.xml      -o live_index.xml   # sha256 e04aeb6a…
curl -sS https://www.kubernetes.dev/blog/index.xml -o live_blog.xml    # sha256 3220efdb…

# 2. local build — Docsy 0.10.0 matters; a stale 0.9.1 node_modules still ships the
#    theme's own list.rss.xml and hides the bug entirely
git checkout fix/restore-rss-template
npm ci
grep -m1 '"version"' node_modules/docsy/package.json          # 0.10.0
rm -rf public && HUGO_ENV=development hugo --environment development

# 3. extract guids and diff
guids() { grep -o '<guid>[^<]*</guid>' "$1" | sed 's|</*guid>||g' \
          | sed 's|https://www.kubernetes.dev/||'; }
guids live_index.xml   > live_index.guids
guids live_blog.xml    > live_blog.guids
guids public/index.xml > local_index.guids

git diff --no-index --color=always live_index.guids local_index.guids | less -R
git diff --no-index --color=always live_blog.guids  local_index.guids | less -R
```

Hugo 0.144.2, matching `netlify.toml`. `git diff --no-index` rather than plain `diff`, since BSD `diff` on macOS has no `--color`.

### `/index.xml` — 752 deletions, 0 insertions

```diff
--- live_index.guids     802 items
+++ local_index.guids     50 items
@@ -1,94 +1,3 @@
-events/2026/kcseu/meet-and-greet/
-events/2026/kcseu/maintainer-summit-eu/
-community/awards/2019/
-community/awards/2020/
-community/awards/2021/
-community/awards/2022/
-community/awards/2023/
-community/awards/2024/
-community/awards/2025/
-resources/calendar/
-events/2020/kcc/how-to-join/
-events/2021/kcc/how-to-join/
-docs/onboarding/01-starting-out/
-resources/services/
-docs/contributor-cheatsheet/
-events/2020/kcc/activities/
-events/2021/kcc/activities/
-community/values/
-events/2022/kcseu/registration/
-events/2022/kcsna/registration/
     ⋮  722 more deletions
-docs/guide/owners/
-docs/guide/expectations/
-docs/orientation/nco-lead-handbook/
-docs/orientation/new-contributor-resources/
-docs/code/cri-api-version-skew-policy/
-docs/code/cri-api-dev-policies/
-resources/release/
-resources/rename/
-community/code-of-conduct/
-community/code-of-conduct-incident-process/
 blog/2026/07/28/kcd-around-the-world-new-york/
 blog/2026/06/26/human-centered-automation-kubernetes-localization-ai-era/
 blog/2026/06/26/open-source-maintainership-in-the-age-of-ai/
 blog/2026/06/24/wg-device-management-spotlight-2026/
 blog/2026/06/15/sig-storage-spotlight-2026/
```

The unprefixed context lines at the bottom are the 50 blog posts, untouched. Deletions by section: `resources/` 602, `events/` 48, `community/` 44, `docs/` 38, `includes/` 2, `search` 1.

733 of the 802 live entries carry `<pubDate>Mon, 01 Jan 0001 00:00:00 +0000</pubDate>`. Zero remain. Feed size drops 3.1 MB → 642 KB.

### `/blog/index.xml` — the 17 archived posts that got auto-reposted

```diff
--- live_blog.guids     67 items
+++ local_index.guids   50 items
@@ -48,20 +48,3 @@
 blog/2022/09/12/k8s-cve-feed-alpha/
 blog/2022/09/09/enhancements-opt-in/
 blog/2022/08/22/sig-storage-spotlight-2022/
-blog/2022/08/15/meet-our-contributors-china-ep-03/
-blog/2022/08/11/enhancing-kubernetes-one-kep-at-a-time/
-blog/2022/08/02/sig-docs-spotlight-2022/
-blog/2022/05/25/contextual-logging/
-blog/2022/05/05/community-meeting-februrary-2022/
-blog/2022/03/15/k8s-triage-bot-helper-ci-job/
-blog/2022/03/14/meet-our-contributors-au-nz-ep-02/
-blog/2022/02/16/sig-node-ci-subproject-celebrates-two-years-of-test-improvements/
-blog/2022/02/04/sig-multicluster-spotlight-2022/
-blog/2022/01/10/meet-our-contributors-india-ep-01/
-blog/2021/12/07/announcing-the-kubernetes-contributor-celebration-2021/
-blog/2021/12/01/improve-your-documentation-with-mermaid.js-diagrams/
-blog/2021/07/09/how-to-choose-a-sig-as-a-non-code-kubernetes-contributor/
-blog/2021/06/01/peeyush-contributor-story/
-blog/2021/05/14/contributor-stories-series/
-blog/2020/09/28/contributing-to-the-development-guide/
-blog/2020/08/24/announcing-the-contributor-website/
```

`@@ -48,20 +48,3 @@` — first 47 items identical, then 20 lines become 3. Item 50 is `sig-storage-spotlight-2022`, the cap boundary. Every deletion is dated 2020–2022, and the last line is the post that went out on X and LinkedIn.

### Channel header

```diff
- <title>Welcome on Kubernetes Contributors</title>
- <description>Recent content in Welcome on Kubernetes Contributors</description>
+ <title>Kubernetes Contributors Blog</title>
+ <description>The Kubernetes contributor blog is used by the project to communicate
+ community reports, contributor news, and any updates that might be relevant to
+ people contributing to Kubernetes.</description>
```

`Welcome` was the home page's `linkTitle` leaking through the old template.

Rebuilt the pre-incident feed by restoring the Docsy 0.9.1 template (`git show 1fb355a:layouts/_default/list.rss.xml`) and reverting `hugo.yaml` and `static/_redirects` to `origin/main`, then diffed its guids against this branch:

```diff
--- pre_incident.guids   50 items
+++ local_index.guids    50 items
```

### Summary

| | live now | after this PR |
|---|---|---|
| feeds generated | 74 | 1 |
| `/index.xml` items | 802 | 50 |
| `01 Jan 0001` entries | 733 | 0 |
| feed size | 3.1 MB | 642 KB |
| `/blog/index.xml` | 67 items | 301 → `/index.xml` |

jeefy (commented):
Looks good, one thing worth double-checking before merge.

**hugo.yaml** — the `outputs` block drops `page: [HTML]`, which means individual pages no longer get any per-page RSS output either (previously theme defaults may have enabled that for some page types). Worth confirming no page templates or partials assume a page-level RSS output format exists (e.g. `.OutputFormats.Get "RSS"` calls elsewhere), otherwise this is fine since your PR explicitly wants only the home feed.

**layouts/index.rss.xml**:
- `{{ .Content | html }}` — should probably be `.Content | safeHTML` (or leave unescaped since `.Content` is already rendered HTML). Using `html` here will double-escape entities in the content and likely mangle the feed body. Worth verifying output actually renders correctly in a feed reader, not just validates as XML.
- The hardcoded `<description>` text is Kubernetes-specific copy carried over from the source template — fine since this is the k8s site, just flagging it's not generic boilerplate in case anyone forks this.
- `site.Config.Services.RSS.Limit` — confirm this key path is correct for the Hugo version pinned in `go.mod`/netlify config; Hugo's config access API has shifted between versions and I want to make sure this doesn't silently return zero-value on the deployed Hugo version (which would still work via the `lt $limit 1` fallback, but good to know it's intentional rather than accidental).

layouts/index.rss.xml
```diff
@@ -0,0 +1,62 @@
...
+      <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
+      {{ with site.Author.email }}<author>{{.}}{{ with site.Author.name }} ({{.}}){{end}}</author>{{end}}
+      <guid>{{ .Permalink }}</guid>
+      <description>
+        {{ $img := (.Resources.ByType "image").GetMatch "*featured*" }}
+        {{ with $img }}
+        {{ $img := .Resize "640x" }}
+        {{ printf "<![CDATA[<img src=\"%s\" width=\"%d\" height=\"%d\"/>]]>" $img.Permalink $img.Width $img.Height | safeHTML }}
+        {{ end }}
+        {{ .Content | html }}
```
TineoC:
```suggestion
        {{ .Content | transform.XMLEscape | safeHTML }}
```

@jeefy I see that `{{ .Content | safeHTML }}` could cause issues for eg

```
input .Content = <p>She said &ldquo;hi&rdquo;</p><img src="a.png">

safeHTML (raw)   FAILS -> undefined entity: line 1, column 36
```

TineoC:
from: https://deploy-preview-855--kubernetes-contributor.netlify.app/blog/2026/07/28/kcd-around-the-world-new-york/

image

Included the RSS button @jeefy 

https://github.com/kubernetes/contributor-site/pull/855/changes#diff-4a7bcb5855bcc928f61527aba2cb16f957266b13ab51553e46016c82ac8c4dd2

Issue #848, opened 2026-08-10
Links inside rendered KEPs point at raw.githubusercontent.com, so they serve unrendered markdown


TineoC:
### What happens

Relative links in a KEP README are rewritten to `raw.githubusercontent.com`, which serves `text/plain`. Clicking a cross-reference inside a rendered KEP page drops the reader into raw markdown source.

Live example — https://www.kubernetes.dev/resources/keps/24/ (AppArmor), the "no perma-Beta requirement" link:

![The 'no perma-Beta requirement' link as rendered on the KEP-24 page](https://raw.githubusercontent.com/TineoC/contributor-site/kep-issue-evidence/.evidence/01-raw-link-source.jpg)

It points to `https://raw.githubusercontent.com/kubernetes/enhancements/master/keps/sig-architecture/1635-prevent-permabeta/README.md`, which renders as:

![Raw markdown source, with HTML comments and unrendered checklists](https://raw.githubusercontent.com/TineoC/contributor-site/kep-issue-evidence/.evidence/02-raw-link-result.jpg)

Same page has two more (`Seccomp`, `Ephemeral Container`), and it affects any KEP whose README uses repo-relative links.

### Where

`content/en/resources/keps/_content.gotmpl` — one base URL is used for both images and links:

```go-html-template
{{- $repoBaseUrl := "https://raw.githubusercontent.com/kubernetes/enhancements/master/" -}}
{{- $kepDirUrl := printf "%skeps/%s/%s/" $repoBaseUrl $kep.owningSig $kep.name -}}
```

`raw.` is right for image sources. For links it should be `blob`.

### Suggested fix

Split the base URL:

```go-html-template
{{- $rawBase  := "https://raw.githubusercontent.com/kubernetes/enhancements/master/" -}}
{{- $blobBase := "https://github.com/kubernetes/enhancements/blob/master/" -}}
{{- $imgDirUrl  := printf "%skeps/%s/%s/" $rawBase  $kep.owningSig $kep.name -}}
{{- $linkDirUrl := printf "%skeps/%s/%s/" $blobBase $kep.owningSig $kep.name -}}
```

Then feed `$imgDirUrl` to the two image rules (`!\[…\]` and `<img src=…`) and `$linkDirUrl` to the two link rules, keeping the current ordering so images are rewritten first.

PR #856, opened 2026-08-12
fix(keps): point markdown links to GitHub blob view instead of raw

Madhurchandran:
## Summary:
Fixes an issue where links inside rendered KEPs point to raw.githubusercontent.com, causing them to display unrendered ra Markdown instead of the rendered GitHub page.

## Changes:
- Update absolute and relative Markdown link rules to use GitHub blob URLs for rendered content.

## Fixes:
#848

### DIFF
--- a/content/en/resources/keps/_content.gotmpl
+++ b/content/en/resources/keps/_content.gotmpl
@@ -108,15 +108,17 @@
         {{- /* Fix relative image and link paths to point to GitHub */ -}}
         {{- $repoBaseUrl := "https://raw.githubusercontent.com/kubernetes/enhancements/master/" -}}
         {{- $kepDirUrl := printf "%skeps/%s/%s/" $repoBaseUrl $kep.owningSig $kep.name -}}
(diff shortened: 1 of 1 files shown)


Priyankasaggu11929:
/assign @TineoC 

for review.

content/en/resources/keps/_content.gotmpl
```diff
@@ -108,15 +108,17 @@
         {{- /* Fix relative image and link paths to point to GitHub */ -}}
         {{- $repoBaseUrl := "https://raw.githubusercontent.com/kubernetes/enhancements/master/" -}}
```
TineoC: is this being unused with this implementation?

Madhurchandran: @TineoC I’ve addressed this in the latest commit by splitting the raw and blob URLs and updating the variable names as suggested in the issue.

TineoC:
Please rebase this branch but the changes looks good so far

Madhurchandran:
Hi @TineoC, it’s rebased now. Would you mind taking a look and assisting with the review when you get a chance?

PR #863, opened 2026-08-19
Publish KCD San Francisco Bay Area blog 

reylejano:
Remove draft status in front matter for the KCD San Francisco Bay Area blog to publish

### DIFF
--- a/content/en/blog/2026/kcd-around-the-world-san-francisco/index.md
+++ b/content/en/blog/2026/kcd-around-the-world-san-francisco/index.md
@@ -1,7 +1,7 @@
 ---
 layout: blog
 title: "KCD Around the World: San Francisco Bay Area"
-draft: true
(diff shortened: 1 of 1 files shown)


TineoC:
Sadly this won't be promoted in social media till the RSS fix is in place in Zappier? Can we hold till 

@chris-short can check this out?

/hold

reylejano:
It would be great to get this blog published since the second KCD SF Bay Area coming up in a little over a week

reylejano:
Is the RSS fix a blocker for this blog to be published?

chris-short:
> Is the RSS fix a blocker for this blog to be published?

No. Not at all. Please proceed with publishing so we can see if the automation works with the updated RSS template.

reylejano:
Thank you @chris-short @TineoC @KashishV999

PR #864, opened 2026-08-19
[Blog] Add Kubernetes v1.37 Declarative Validation blog post

lalitc375:
Add Kubernetes v1.37 Declarative Validation blog post

### Content Overview
Authored by Lalit Chauhan and Yongrui Lin, this post highlights the massive adoption of Declarative Validation (DV) in Kubernetes v1.37. Key points include:
* A record-breaking 118 API reviews were completed in v1.37.
* Unique field adoption for Declarative Validation grew by 107.7% compared to v1.36.
* How the shift from legacy handwritten validation to structural tags (e.g., `+k8s:maxLength=64`) saves reviewer time and enforces API correctness.
* The evolution of `validation-gen` and the introduction of 6 new validation tags.

PR #870, opened 2026-08-27
Fix API review metrics link to point to public project board

liggitt:
cc @lalitc375 @lmktfy

addresses https://github.com/kubernetes/contributor-site/pull/864/changes#r3855260463


yongruilin:
/lgtm
Thanks @liggitt

PR #871, opened 2026-08-28
Add SRI hash to DataTables CSS link

### LINKED ISSUE #851: KEP index loads jQuery and DataTables from third-party CDNs; the DataTables stylesheet has no SRI
### What happens

Loading https://www.kubernetes.dev/resources/keps/ pulls runtime assets from origins outside the project:

| Request | Origin | SRI |
|---|---|---|
| `jquery-3.6.0.min.js` | `code.jquery.com` | yes |
| `1.12.1/js/jquery.dataTables.js` | `cdn.datatables.net` | yes |
| `1.12.1/css/jquery.dataTables.css` | `cdn.datatables.net` | **no** |

Captured from the network log on a plain page load today.

### Problems

- The stylesheet has no `integrity` attribute, so it is the one asset here with nothing pinning its contents.
- DataTables is pinned to 1.12.1, a 2022 release, with nothing in the repo tracking it.
- If the CDN is blocked or unavailable, `body-end.html` returns early at `if (typeof $ === 'undefined' || !$.fn || !$.fn.DataTable) return;` and the page renders every KEP as one unpaginated table (see #6 below — currently 640 rows).
- Visitors' requests reach two third-party hosts to view a Kubernetes project page.

### Where

`layouts/partials/hooks/body-end.html` for the DataTables script; the jQuery and DataTables CSS tags come from the site's head config.

### Suggested fix

Vendor the assets and let Hugo fingerprint them:

```go-html-template
{{ $dt := resources.Get "js/jquery.dataTables.js" | fingerprint }}
<script src="{{ $dt.RelPermalink }}" integrity="{{ $dt.Data.Integrity }}"></script>
```

magic-peach:
## What type of PR is this?
/kind bug

The DataTables CSS loaded from the CDN in `layouts/partials/hooks/head-end.html` was missing a Subresource Integrity (SRI) attribute, unlike the DataTables JS file loaded elsewhere on pages using the `keps-data` shortcode. Adds `integrity` and `crossorigin` attributes, with the hash computed against the exact file served from the pinned `1.12.1` CDN URL.

Fixes #851

#### AI Disclosure
PR description was written using AI  

## Special notes for reviewer:
Verified the hash by downloading the exact CDN file and computing sha384 directly, and cross-checked the method by reproducing the already-present JS file's hash byte-for-byte from the same CDN first.


lmktfy:
> Verified the hash by downloading the exact CDN file and computing sha384 directly, and cross-checked the method by reproducing the already-present JS file's hash byte-for-byte from the same CDN first.

This isn't the right check to have done, so far as I can think it through.

skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

PR #874, opened 2026-08-28
Exclude template/unnumbered KEPs from rendering

### LINKED ISSUE #847: KEP template and unnumbered KEPs render as real KEPs at /resources/keps/0000/
### What happened

https://www.kubernetes.dev/resources/keps/0000/ renders as a normal KEP page. `0000` is the placeholder KEP number from the template, not an assigned one.

Four directories in `kubernetes/enhancements` still have `kep-number: 0000`:

| Path | Real KEP? |
|---|---|
| `keps/sig-architecture/0000-kep-process` | yes |
| `keps/sig-release/0000-anago-to-krel-migration` | yes |
| `keps/sig-contributor-experience/0000-community-forum` | yes |
| `keps/sig-cloud-provider/providers/0000-cloud-provider-template` | **template** (author `janedoe`, creation date `yyyy-mm-dd`) |

Because the generated page path is the KEP number, all four resolve to the same URL and overwrite each other — the page currently shows sig-release's "Anago to Krel Migration". So three legitimate KEPs are unreachable, and a template is published alongside accepted enhancements.

### Where

`content/en/resources/keps/_content.gotmpl`:

```go-html-template
{{/* Build page path: /resources/keps/{kepNumber}/ */}}
{{- $pagePath := $kep.kepNumber -}}
```

There is no uniqueness check, so the last KEP in `keps.json` with a given number silently wins.

The root `keps/NNNN-kep-template/kep.yaml` is already excluded, but only incidentally: its `stage`/`status` are literal placeholder strings (`alpha|beta|stable`) that fail the stage/status filter.

### Expected

- Templates do not appear in the KEP index or get a detail page.
- KEPs without an assigned number either do not publish or publish at distinct URLs.

### Suggested fix

1. Skip entries whose `kepNumber` is not a real assigned number (`0000`, `NNNN`) or whose `name` ends in `-template`, with a `warnf` so skips are visible in the build log.
2. Optionally make the page path collision-proof (e.g. `{owningSig}/{name}`), so a duplicated number never silently drops a KEP again.

cc. @lmktfy

magic-peach:
## What type of PR is this?
/kind bug

KEP directories that still carry a placeholder `kepNumber` (`0000`, `NNNN`) -- i.e. template or process-doc directories, not real KEPs -- were rendering as if they were real KEPs. Multiple such entries also collided onto the same `/resources/keps/0000/` page path. This skips entries with a placeholder `kepNumber`, and adds a page-path uniqueness guard as defense-in-depth against any other future collision.

Fixes #847

## AI Disclosure
Wrote the PR description and pushed it using AI
 
## Special notes for your reviewer:
The issue also suggested excluding entries whose directory name ends in `-template`. I checked the live KEP data and found KEP 5978 (`cluster-resource-claim-template`, sig-node) is a real, numbered, currently active KEP whose name happens to end in `-template` because the feature itself is a `ClusterResourceClaimTemplate` resource -- filtering by name would have wrongly dropped it. All 4 problem entries in the issue already share `kepNumber: "0000"`, so filtering on that alone fully fixes the bug without the regression. Filtering by name is intentionally not implemented here.

Verified via build log: all 4 "0000" entries (sig-architecture/0000-kep-process, sig-cloud-provider/0000-cloud-provider-template, sig-contributor-experience/0000-community-forum, sig-release/0000-anago-to-krel-migration) are now skipped with a warning, `/resources/keps/0000/` no longer exists, and KEP 5978 still publishes normally. Total page count dropped from 850 to 849, matching the expected "4 entries that used to collide onto 1 page -> 0 pages."


lmktfy:
You haven't disclosed AI use, and yet I believe you would have used AI help here. Have I got that right?

magic-peach:
I used ai to write the pr description

skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

PR #882, opened 2026-09-11
fix: preserve KEP search in shareable URLs

### LINKED ISSUE #853: KEP index search term is missing from the shareable URL
### What happens

The KEP index syncs release, SIG, and stage filters into the URL hash, but not the search box. Share the URL after searching and the recipient gets your filters with an empty search.

Repro on https://www.kubernetes.dev/resources/keps/ :

1. Set SIG to `sig-node`.
2. Type `swap` in the search box.
3. Table narrows to 1 row.
4. URL reads `.../resources/keps/#sig=sig-node` — no trace of the search term.

### Example

Image

https://www.kubernetes.dev/resources/keps/#sig=sig-node

### Where

`layouts/partials/hooks/body-end.html`. `syncHash()` writes `release`, `sig`, and `stage`; the search input's listener only calls `dt.draw()`:

```js
searchInput.addEventListener('input', function() {
  searchTerm = this.value;
  dt.draw();
});
```

### Suggested fix

Add the term to both halves. In `syncHash()`:

```js
if (searchTerm) parts.push('q=' + encodeURIComponent(searchTerm));
```

and in `applyHash()`, restore `searchTerm` and the input's `value` from `params.q` before the `dt.draw()` that is already there. Worth debouncing the input listener's `syncHash()` call so it isn't one `history.replaceState` per keystroke.

immanuwell:
Fixes #853

The KEP search filters the table, but never reaches the URL. 
Shared links silently drop the term

This fix stores search as `q`, restores it on load, and debounces URL updates

Repro:
1. Open `https://www.kubernetes.dev/resources/keps/`
2. Select `sig-node`
3. Search for `swap`
4. Copy and reload the URL. The search is gone and 126 rows return instead of 1

### DIFF
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -33,6 +33,7 @@
     var $table = $(table);
     var activeStages = new Set();
     var searchTerm = '';
+    var searchHashTimer;
 
     // Cache elements outside the search loop callback to prevent performance issues
     var selRelease = document.getElementById('kep-filter-release');
@@ -102,6 +103,8 @@
       searchInput.addEventListener('input', function() {
         searchTerm = this.value;
         dt.draw();
+        clearTimeout(searchHashTimer);
+        searchHashTimer = setTimeout(syncHash, 200);
       });
     }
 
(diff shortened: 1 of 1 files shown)


layouts/partials/hooks/body-end.html
```diff
@@ -156,6 +159,7 @@
       if (r) parts.push('release=' + encodeURIComponent(r));
       if (s) parts.push('sig=' + encodeURIComponent(s));
       if (activeStages.size) parts.push('stage=' + encodeURIComponent(Array.from(activeStages).join(',')));
+      if (searchTerm) parts.push('q=' + encodeURIComponent(searchTerm));
       var hash = parts.length ? '#' + parts.join('&') : '';
       history.replaceState(null, '', hash || window.location.pathname);
```
lmktfy, line 164: Rather than _replace_, I would wait for an Enter keypress and then append a new URL to history. That's similar to how Google, etc work.

